PR #3, opened 2020-03-31
netaddr: implement IP.IPAddr for interop with existing code

### LINKED ISSUE #2: proposal: add method IP.IPAddr for adapting into existing code
I am certainly a fan of the ideas being put forth by this package, but there is a massive amount of preexisting code which expects to work with net.IP/net.IPAddr exclusively.

I'd propose a method as follows:

```go
// IP returns the net.IPAddr representation of an IP. If the IP contains a zone
// identifier, the Zone field of the returned IPAddr is populated.
func (ip IP) IPAddr() *net.IPAddr
```

From there, the caller can either use the net.IPAddr directly or extract the IP address for use with APIs which expect net.IP.

I was considering proposing `IP() net.IP` instead, but due to the inclusion of the zone identifier in this package, I think this is a fair compromise.

mdlayher:

Thoughts welcome on the "what to do with nil" TODO. I am tempted to panic to avoid an error return and also a potential nil pointer dereference in calling code. But I'm open to all suggestions.

Fixes #2.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -86,6 +86,37 @@ func ParseIP(s string) (IP, error) {
 	return IP{v6}, nil
 }
 
+// IPAddr returns the net.IPAddr representation of an IP. The returned value is
+// always non-nil, but the IPAddr.IP will be nil if ip is the zero value.
+// If ip contains a zone identifier, IPAddr.Zone is populated.
+func (ip IP) IPAddr() *net.IPAddr {
+	switch ip := ip.ipImpl.(type) {
+	case nil:
+		// Nothing to do.
+		return &net.IPAddr{}
+	case v4Addr:
+		return &net.IPAddr{
+			// Assume the caller wants the 4-byte representation.
+			IP: net.IPv4(ip[0], ip[1], ip[2], ip[3]).To4(),
+		}
+	case v6Addr:
+		b := make(net.IP, net.IPv6len)
+		copy(b, ip[:])
+
+		return &net.IPAddr{IP: b}
+	case v6AddrZone:
+		b := make(net.IP, net.IPv6len)
+		copy(b, ip.v6Addr[:])
+
+		return &net.IPAddr{
+			IP:   b,
+			Zone: ip.zone,
+		}
+	default:
+		panic("netaddr: unhandled ipImpl representation")
+	}
+}
+
 // Is4 reports whether ip is an IPv4 address.
 //
 // TODO: decide/clarify the behavior for IPv4-mapped IPv6 addresses.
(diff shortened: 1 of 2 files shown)


bradfitz:
Case nil: return &net.IPAddr{} ?

mdlayher:
I'm good with that. Can you do anything useful with that value though?

bradfitz:
Not crashing is useful! They can use the ipa.IP safely without checking the returned ipa being nil, and then they'll just get an error from the thing that took a nil net.IP.

netaddr.go
```diff
@@ -86,6 +86,38 @@ func ParseIP(s string) (IP, error) {
 	return IP{v6}, nil
 }
 
+// IPAddr returns the net.IPAddr representation of an IP. If the IP contains a
+// zone identifier, the Zone field of the returned IPAddr is populated.
+func (ip IP) IPAddr() *net.IPAddr {
+	if ip.ipImpl == nil {
```
bradfitz: just put this in the switch below as the first case (case nil ... https://play.golang.org/p/5fMIvT2zg35)

netaddr.go
```diff
@@ -86,6 +86,38 @@ func ParseIP(s string) (IP, error) {
 	return IP{v6}, nil
 }
 
+// IPAddr returns the net.IPAddr representation of an IP. If the IP contains a
+// zone identifier, the Zone field of the returned IPAddr is populated.
```
bradfitz: document that the returned value is always non-nil, but the IPAddr.IP will be nil if ip is the zero value?

mdlayher:
All done, one more review?

netaddr.go
```diff
@@ -86,6 +86,37 @@ func ParseIP(s string) (IP, error) {
...
+// always non-nil, but the IPAddr.IP will be nil if ip is the zero value.
+// If ip contains a zone identifier, IPAddr.Zone is populated.
+func (ip IP) IPAddr() *net.IPAddr {
+	switch ip := ip.ipImpl.(type) {
+	case nil:
+		// Nothing to do.
+		return &net.IPAddr{}
+	case v4Addr:
+		return &net.IPAddr{
+			// Assume the caller wants the 4-byte representation.
```
bradfitz: Then why are you returning the 16 byte form?

https://golang.org/pkg/net/#IPv4 says "IPv4 returns the IP address (in 16-byte form) of the IPv4 address a.b.c.d."

netaddr.go
```diff
@@ -86,6 +86,37 @@ func ParseIP(s string) (IP, error) {
...
+			IP: net.IPv4(ip[0], ip[1], ip[2], ip[3]).To4(),
```
bradfitz: That's a little indirect, but I guess.

We could also simplify it in a follow-up change to just return the 4 bytes.

PR #4, opened 2020-03-31
netaddr: simplify IPv4 IPAddr conversion

mdlayher:

A nice simplification as suggested by @bradfitz in #3.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -95,10 +95,7 @@ func (ip IP) IPAddr() *net.IPAddr {
 		// Nothing to do.
 		return &net.IPAddr{}
 	case v4Addr:
-		return &net.IPAddr{
-			// Assume the caller wants the 4-byte representation.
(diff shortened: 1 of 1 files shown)


netaddr.go
```diff
@@ -95,10 +95,10 @@ func (ip IP) IPAddr() *net.IPAddr {
...
 		return &net.IPAddr{}
 	case v4Addr:
-		return &net.IPAddr{
-			// Assume the caller wants the 4-byte representation.
-			IP: net.IPv4(ip[0], ip[1], ip[2], ip[3]).To4(),
-		}
+		b := make(net.IP, net.IPv4len)
+		copy(b, ip[:])
+
+		return &net.IPAddr{IP: b}
```
bradfitz: or even just `return &net.IPAddr{IP: net.IP{ip[0], ip[1], ip[2], ip[3]}}` ?

mdlayher: Yep makes sense.

PR #12, opened 2020-04-13
netaddr: add IPPrefix type

crawshaw:


netaddr.go
```diff
@@ -323,3 +325,94 @@ func (p IPPort) TCPAddr() *net.TCPAddr {
...
+// The range of Bits is [0,32] for IPv4 or [0,128] for IPv6.
+type IPPrefix struct {
+	IP   IP
+	Bits uint8
+}
+
+// ParseIPPrefix parses s as an IP address prefix.
+// The string can be in the form "192.168.1.0/24" or "2001::db8::/32".
+func ParseIPPrefix(s string) (IPPrefix, error) {
+	i := strings.IndexByte(s, '/')
```
mdlayher, line 341: strings.Split and check len == 2 to simplify?

netaddr.go
```diff
@@ -323,3 +325,94 @@ func (p IPPort) TCPAddr() *net.TCPAddr {
...
+// The first Bits of IP are specified, the remaining bits match any address.
+// The range of Bits is [0,32] for IPv4 or [0,128] for IPv6.
+type IPPrefix struct {
+	IP   IP
+	Bits uint8
+}
+
+// ParseIPPrefix parses s as an IP address prefix.
+// The string can be in the form "192.168.1.0/24" or "2001::db8::/32".
+func ParseIPPrefix(s string) (IPPrefix, error) {
```
mdlayher, line 340: Has it been decided that everything in this package should be passed around without pointers?

netaddr_test.go
```diff
@@ -272,6 +272,65 @@ func TestIs4In6(t *testing.T) {
 	}
 }
 
+func TestIPPrefix(t *testing.T) {
```
mdlayher, line 275: It'd be nice to see some test cases for erroneous inputs too.

netaddr_test.go
```diff
@@ -281,6 +340,13 @@ func mustIP(s string) IP {
 	return ip
 }
 
+func mustIPs(strs ...string) (res []IP) {
```
mdlayher, line 343: Nit: I'm not sure the named return is necessary for a function with a single return.

netaddr.go
mdlayher, line 340: I think either the doc comment or the name should include something about CIDR notation.

netaddr.go
bradfitz: We don't want the strings.Split allocation.

netaddr.go
bradfitz: Where possible/small, yes.

netaddr.go
```diff
@@ -323,3 +325,94 @@ func (p IPPort) TCPAddr() *net.TCPAddr {
...
+		bits = 32
+	}
+	stdIP, _ := p.IP.ipZone()
+	return &net.IPNet{
+		IP:   stdIP,
+		Mask: net.CIDRMask(int(p.Bits), bits),
+	}
+}
+
+// Contains returns true if the network includes ip.
```
bradfitz, line 382: reports whether

netaddr.go
bradfitz: (We don't want the allocation)

netaddr.go
```diff
@@ -323,3 +325,94 @@ func (p IPPort) TCPAddr() *net.TCPAddr {
 		Zone: zone,
 	}
 }
+
+// IPPrefix is an IP address prefix representing an IP network.
+//
+// The first Bits of IP are specified, the remaining bits match any address.
+// The range of Bits is [0,32] for IPv4 or [0,128] for IPv6.
```
bradfitz, line 332: What's the meaning of 0 for bits? Move the Bits docs to the field below?

netaddr_test.go
```diff
@@ -357,3 +423,15 @@ func BenchmarkIPv6(b *testing.B) {
 		}
 	}
 }
+
+func BenchmarkIPv4Contains(b *testing.B) {
+	b.ReportAllocs()
```
bradfitz, line 428: also add a test that checks allocs (testing.AllocsPerRun) to make sure it's 0 else fails?

netaddr.go
bradfitz: document that this a 4-in-6 IP is not in a matching IPPrefix with a v4 IP?

netaddr.go
mdlayher: Ack, and +1 to adding a test that verifies no allocations. I would also consider leaving a comment stating that goal, but that's less necessary once a test is added.

netaddr.go
mdlayher: /0 can be used to indicate the entire address space, e.g. `0.0.0.0/0` and `::/0`. It's unusual but valid.

bradfitz: Test for that then.

crawshaw:
These commits can all be squashed into the original before merging, just experimenting to see if leaving them unsquashed on my end improves the github PR experience for reviewers.

netaddr_test.go
```diff
@@ -272,6 +273,135 @@ func TestIs4In6(t *testing.T) {
...
+			}
+			if got := prefix.String(); got != test.prefix {
+				t.Errorf("prefix.String()=%q, want %q", got, test.prefix)
+			}
+		})
+	}
+}
+
+func TestParseIPPrefixAllocs(t *testing.T) {
+	ip := "192.168.1.0"
```
mdlayher, line 348: Nit: test IPv6 allocations as well for completeness?

crawshaw: seems entirely unnecessary given the design of the test, but sure. done.

netaddr.go
```diff
@@ -323,3 +325,98 @@ func (p IPPort) TCPAddr() *net.TCPAddr {
...
+		bits = 32
+	}
+	stdIP, _ := p.IP.ipZone()
+	return &net.IPNet{
+		IP:   stdIP,
+		Mask: net.CIDRMask(int(p.Bits), bits),
+	}
+}
+
+// Contains reports whether the network includes addr.
```
bradfitz, line 383: "the network" here to mean the IPPrefix is a bit odd, even if if the godoc on IPPrefix calls it a network.

Maybe at least say "the network p includes addr" ?

netaddr_test.go
mdlayher: Thanks! For my own curiosity: what makes it unnecessary?

I see that ParseIPPrefix uses a single code path for v4/v6, but I do see a slight divergence for ParseIP due to the different underlying types. I figured it would make sense to verify both are zero allocation, but perhaps I misunderstand?

crawshaw: The test as written runs both ParseIP and ParseIPPrefix. It subtracts the allocs of the former from the latter. So if ParseIP goes and generates N more allocs for IPv6, it makes no difference to the accounting here.

Besides the ParseIP which we are explicitly subtracting, there are no divergent codepaths for v4 c.f. v6 so adding a v6 address does not increase effective coverage.

This is an example of where a code coverage tool would generate an incorrect report, because it can't see the test is annulling a lot of code paths by doing `allocs(ParseIPPrefix)-allocs(ParseIP)`.

mdlayher: Understood, thanks for the follow up!

PR #14, opened 2020-04-15
ci: set up GitHub actions workflows

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

Mostly copied from tailscale repo. I think this will work, but this is my first time setting up Actions.

Fixes #13.

PR #15, opened 2020-04-15
netaddr: fix staticcheck issues identified in #14

mdlayher:

Make the linter happy. The file-ignore directive isn't ideal but it's probably okay for tests.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -158,10 +158,10 @@ func (ip IP) Less(ip2 IP) bool {
 	switch bytes.Compare(a16[:], b16[:]) {
 	case -1:
 		return true
-	default:
-		return a.Zone() < b.Zone()
 	case 1:
 		return false
+	default: // case 0 (bytes.Compare only returns -1, 1, 0)
+		return a.Zone() < b.Zone()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


netaddr.go
```diff
@@ -158,10 +158,12 @@ func (ip IP) Less(ip2 IP) bool {
 	switch bytes.Compare(a16[:], b16[:]) {
 	case -1:
 		return true
-	default:
+	case 0:
```
bradfitz, line 161: what was wrong with this?

mibk: See https://github.com/dominikh/go-tools/issues/396. It seems as a weird check to me as well.

mdlayher: This was slightly surprising to me as well. I think on the whole that staticcheck makes good suggestions and we could add a directive to ignore this one though.

Although the behavior of bytes.Compare is guaranteed by the stdlib, I've typically taken a stance of using panics in default cases for impossible behavior like `bytes.Compare(a, b) == 2`.

bradfitz: I'd rather not have a panic case here on principle. Plus it's slightly faster to not have a 4th case. (probably barely measurable, but...)

But I guess we can arrange it. Just add a comment:

```
    default: // case 0 (bytes.Compare only returns -1, 1, 0)
```

mdlayher: Sure, done.

PR #16, opened 2020-04-16
netaddr: implement IP.Prefix to produce IPPrefix

### LINKED ISSUE #9: proposal: netaddr.Mask type for IP address masking
I was planning to propose netaddr.Prefix, but I think netaddr.Mask is a prerequisite.

I was taking a look at net.IPMask and wondered why it was represented with []byte instead of an integer, and I can only assume it is to support the insanity of IPv4 non-contiguous subnet masks.

I am wondering whether it's even necessary to indicate an IP address family in a mask as I suspect the caller would always know if they're working with IPv4 or IPv6 addresses, so perhaps something as minimal as this would work:

```go
// An IPMask is a CIDR IP address mask length value.
// TODO: consider a struct instead to hide internal representation? Maybe to disallow
// value >128?
type IPMask byte

// IPv4Mask returns the IP mask (in 4-byte form) of the IPv4 mask a.b.c.d. The input
// bytes must contain all leading ones followed by all trailing zeros or
// IPv4Mask will return an error.
func IPv4Mask(a, b, c, d byte) (IPMask, error) {}
```

And then a method to mask IP addresses identical to what stdlib does:

```go
// Mask returns the result of masking the IP address ip with mask.
func (ip IP) Mask(mask IPMask) IP {}
```

mdlayher:

This is my first shot and feedback is welcome. I left TODOs for things I'd like feedback on.

For allocations, at the moment we have:

```
./netaddr.go:289:13: make([]byte, len(b)) escapes to heap
./netaddr.go:310:12: v4 escapes to heap                                                  
./netaddr.go:314:12: v6 escapes to heap                                                  
./netaddr.go:320:12: v6 escapes to heap 
```

Fixes #9 .

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -327,6 +327,73 @@ func (ip IP) IsMulticast() bool {
 	}
 }
 
+// Prefix applies a CIDR mask of leading bits to IP, producing an IPPrefix
+// of the specified length. If IP is the zero value, a zero-value IPPrefix and
+// a nil error are returned. If bits is larger than 32 for an IPv4 address or
+// 128 for an IPv6 address, an error is returned.
+func (ip IP) Prefix(bits uint8) (IPPrefix, error) {
+	maxBits := uint8(32)
+	if ip.Is6() {
+		maxBits = 128
+	}
+
+	if bits > maxBits {
+		return IPPrefix{}, fmt.Errorf("netaddr: prefix length %d too large for IP address family", bits)
+	}
+
+	var b []byte
+	switch ip := ip.ipImpl.(type) {
+	case nil:
+		// Zero-value input produces zero-value output.
+		return IPPrefix{}, nil
+	case v4Addr:
+		b = ip[:]
+	case v6Addr:
+		b = ip[:]
+	case v6AddrZone:
+		b = ip.v6Addr[:]
+	default:
+		panic("netaddr: unhandled ipImpl representation")
+	}
+
+	// Apply the mask specified by bits.
+	n := bits
+	for i := 0; i < len(b); i++ {
+		if n >= 8 {
+			b[i] &= 0xff
+			n -= 8
+			continue
+		}
+
+		b[i] = ^byte(0xff >> n)
+		n = 0
+	}
+
+	var out IP
+	switch ip.ipImpl.(type) {
+	case v4Addr:
+		var v4 v4Addr
+		copy(v4[:], b)
+		out = IP{ipImpl: v4}
+	case v6Addr:
+		var v6 v6Addr
+		copy(v6[:], b)
+		out = IP{ipImpl: v6}
+	case v6AddrZone:
+		var v6 v6AddrZone
+		copy(v6.v6Addr[:], b)
+		v6.zone = ip.Zone()
+		out = IP{ipImpl: v6}
+	default:
+		panic("netaddr: unhandled ipImpl representation")
+	}
+
+	return IPPrefix{
+		IP:   out,
+		Bits: bits,
+	}, nil
+}
+
 // String returns the string form of the IP address ip.
 // It returns one of 4 forms:
 //
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -352,6 +352,119 @@ func TestLess(t *testing.T) {
 	}
 }
 
+func TestIPPrefixMasking(t *testing.T) {
+	type subtest struct {
+		ip   IP
+		bits uint8
+		p    IPPrefix
+		ok   bool
+	}
+
+	// makeIPv6 produces a set of IPv6 subtests with an optional zone identifier.
+	makeIPv6 := func(zone string) []subtest {
+		if zone != "" {
+			zone = "%" + zone
+		}
+
+		return []subtest{
+			{
+				ip:   mustIP(fmt.Sprintf("2001:db8::1%s", zone)),
+				bits: 255,
+			},
+			{
+				ip:   mustIP(fmt.Sprintf("2001:db8::1%s", zone)),
+				bits: 32,
+				p:    mustIPPrefix(fmt.Sprintf("2001:db8::%s/32", zone)),
+				ok:   true,
+			},
+			{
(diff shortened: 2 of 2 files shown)


mdlayher:
Sweet, just realized I can apparently remove the temporary buffer due to IP.Mask using a value receiver.

```
./netaddr.go:258:7: leaking param: ip                                                    
./netaddr.go:302:12: v4 escapes to heap                                                  
./netaddr.go:306:12: v6 escapes to heap                                                  
./netaddr.go:312:12: v6 escapes to heap  
```

netaddr.go
```diff
@@ -252,6 +252,75 @@ func (ip IP) IsMulticast() bool {
...
+		// TODO: does it makes sense to apply zone in an IPPrefix?
+		var v6 v6AddrZone
+		copy(v6.v6Addr[:], b)
+		v6.zone = ip.Zone()
+		out = IP{ipImpl: v6}
+	default:
+		panic("netaddr: unhandled ipImpl representation")
+	}
+
+	// TODO: return bare IP instead?
```
bradfitz, line 317: perhaps? MaskPrefix could then be a couple line wrapper that returns IPPrefix isntead to save the caller a few lines?

@danderson, @crawshaw?

danderson: Off the top of my head, I can't think why I would want to mask an IP for reasons other than constructing a prefix. Maybe if I want to do some kind of enumeration of the network, or construct one IP from another?

Having Mask return a bare IP seems fine if someone has a use in mind for it. I can't think of one.

netaddr.go
```diff
@@ -252,6 +252,75 @@ func (ip IP) IsMulticast() bool {
...
+	case v4Addr:
+		var v4 v4Addr
+		copy(v4[:], b)
+		out = IP{ipImpl: v4}
+	case v6Addr:
+		var v6 v6Addr
+		copy(v6[:], b)
+		out = IP{ipImpl: v6}
+	case v6AddrZone:
+		// TODO: does it makes sense to apply zone in an IPPrefix?
```
danderson, line 308: It's unusual, but I don't see a reason to make it not-work. If only so that you don't have to treat zoned v6 addrs specially in generic code, unless you specifically care.

netaddr.go
```diff
@@ -252,6 +252,75 @@ func (ip IP) IsMulticast() bool {
...
+// zero-value IPPrefix is returned.
+func (ip IP) Mask(bits uint8) IPPrefix {
+	maxBits := uint8(32)
+	if ip.Is6() {
+		maxBits = 128
+	}
+
+	if bits > maxBits {
+		// Invalid mask size for both IPv4/6: nothing to do.
+		return IPPrefix{}
```
danderson, line 266: Does this square with the rest of the API? Returning zero values on error is a very stdlib net kind of thing to do, but in other PRs I thought I saw that we wanted to use the standard `ret, err` pattern?

netaddr.go
mdlayher: I would generally agree you want the prefix if you are masking. But it's trivial to construct an  IPPrefix yourself since you already had to specify the input bits anyway:

```go
const bits = 64
masked := ip.Mask(bits)
p := netaddr.Prefix{IP: masked, Bits: bits}
```

netaddr.go
bradfitz: I'm less bothered here because the invalid input is unlikely to come from users/network. It's just a number.

But I could go either way.

netaddr.go
mdlayher: I think I'm in favor of doing this as well actually. I had thought there were some other zero-value returns for invalid input, but it seems that most of those are around handling the zero-value IP.

For anything that accepts input, it is probably best to return errors.

mdlayher: Before I change anything: assuming I add an error return, what's the proper behavior for `case nil` below? `return IPPrefix{}, nil` zero value still?

mdlayher: @bradfitz was https://github.com/inetaf/netaddr/pull/16#discussion_r410464239 in response to this discussion?

netaddr.go
crawshaw: Similarly it's very easy to extract just the IP from the returned prefix:

```
masked := ip.Mask(bits).IP
```

Given there's no functionality difference and no obvious performance difference (I would expect an optimizing compiler to inline this function then remove the Prefix creation in the above example), I say go with the return type that's more common. In this case I suspect it's the whole IPPrefix.

netaddr.go
crawshaw: An alternative would be to panic on invalid input to the function. I don't see how an invalid bits value is an expected error, it feels to me like it would always be a programming error.

bradfitz:
If it returns Prefix I'd prefer the method be named Prefix, not Mask.

netaddr.go
danderson: I could see passing in a bits value that's based on user input. Panicking in that case is correct, but rude.

I'm not particularly fussed about zero value vs. errors as the return, as long as there's an easy way to check for errors (e.g. something like time.IsZero(), if we have one of those). But if a function can fail, barring extenuating circumstances I'd rather have it return errors.

netaddr.go
mdlayher: Ack, will do.

netaddr.go
mdlayher: Alright, will add error returns momentarily and ask for more feedback.

mdlayher:
SGTM on the rename.

netaddr.go
```diff
@@ -305,6 +305,73 @@ func (ip IP) IsMulticast() bool {
 	}
 }
 
+// Prefix applies a CIDR mask of leading bits to IP, producing an IPPrefix
+// of the specified length. If IP is the zero value, a zero-value IPPrefix is
```
bradfitz, line 309: > If IP is the zero value, ...

It's not exactly clear whether this sentence means _with a nil error_ now. Maybe just add `(and a nil error)`.

mdlayher:
Done, rebased on master, and squashed into one commit. Ready for another look, @bradfitz .

PR #18, opened 2020-04-17
netaddr: add IP.IsLinkLocalUnicast method

mdlayher:

With this and #16, all of my CoreRAD needs have been met by netaddr. I also slightly adjusted the unicast fixtures to be more correct and not overlap with link-local.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -234,6 +234,24 @@ func (ip IP) Unmap() IP {
 	return IP{v4Addr{a[12], a[13], a[14], a[15]}}
 }
 
+// IsLinkLocalUnicast reports whether ip is a link-local unicast address.
+// If ip is the zero value, it will return false.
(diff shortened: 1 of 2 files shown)


netaddr.go
```diff
@@ -234,6 +234,24 @@ func (ip IP) Unmap() IP {
 	return IP{v4Addr{a[12], a[13], a[14], a[15]}}
 }
 
+// IsLinkLocalUnicast reports whether ip is a link-local unicast address. If ip
```
bradfitz, line 237: Start "If ip" sentence on its own line for readers reading source instead of godoc elsewhere.

mdlayher: Done.

mdlayher:
Ah shoot, I created a merge commit out of habit. Feel free to rewrite master to remove it if you'd like.

PR #22, opened 2020-04-20
ci: use build matrix to test with multiple Go versions

tklauser:
Use a build matrix to test with Go 1.11, 1.12, 1.13 and Go 1.14.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -10,13 +10,16 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        go-version: [1.11, 1.12, 1.13, 1.14]
     runs-on: ubuntu-latest
 
     steps:
     - name: Set up Go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.14
+        go-version: ${{ go-version }}
       id: go
 
     - name: Check out code into the Go module directory


.github/workflows/build-and-test.yml
```diff
@@ -10,7 +10,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: [1.13, 1.14]
+        platform: [ubuntu-latest, macos-latest]
```
bradfitz, line 16: There's nothing non-portable so I don't think we need to test on macOS. (It's pretty slow on GitHub actions... I'd rather only use it if it buys us something.)

If anything, I'd prefer to add older Go versions and stay on Linux. I think this is worth supporting back to Go 1.11 or Go 1.12.

tklauser: Good point. Will drop the macOS part and focus on testing with older Go versions. Also noticed, that I actually didn't change the action to use the Go version from the build matrix :facepalm: PR now updated.

.github/workflows/linux.yml
```diff
@@ -10,13 +10,16 @@ on:
...
+      matrix:
+        go-version: [1.11, 1.12, 1.13, 1.14]
     runs-on: ubuntu-latest
 
     steps:
     - name: Set up Go
       uses: actions/setup-go@v1
       with:
-        go-version: 1.14
+        go-version: ${{ go-version }}
```
tklauser, line 22: @bradfitz Sorry, just noticed that this should have been `${{ matrix.go-version }}` and the action thus didn't run. Will send a follow-up PR.

tklauser: #23

PR #23, opened 2020-04-20
ci: fix matrix property name after #22

tklauser:
Use the correct property matrix.go-version to fix the Github action and
use the Go version defined in the build matrix.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v1
       with:
-        go-version: ${{ go-version }}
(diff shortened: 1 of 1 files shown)


tklauser:
@bradfitz looks like this doesn't build against Go < 1.13 due to the use of `%w` error wrapping. What would you suggest? Disable Go 1.11 and 1.12 for now? Don't use error wrapping? Let it fail for now (it seems that GH actions will cancel the 1.13 an 1.14 builds though)?

bradfitz:
Just pushed a change to remove use of %w for now. It wasn't buying us much yet. We can do some +build conditional wrapping later if we find a more valuable use of %w.

tklauser:
Thanks

PR #24, opened 2020-04-21
netaddr: add FromStdIPNet to convert *net.IPNet to IPPrefix

mdlayher:

Mirrors the existing From* APIs.


mdlayher:
No rush as I can get by with `ParseIPPrefix(cidr.String())` too, but let me know what you all think of this. Thanks!

PR #25, opened 2020-04-25
netaddr: add initial well-known IP addresses

mdlayher:

These are the only two "constants" I need at the moment. The standard library exposes a few more.

It'd be nice if these could be immutable but I think this is probably reasonable.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -97,6 +97,22 @@ func (ip v6AddrZone) String() string {
 	return (&net.IPAddr{IP: net.IP(ip.v6Addr[:]), Zone: ip.zone}).String()
 }
 
+// Well known IP addresses which are accessed only through exported functions,
+// so the linker can eliminate them if they are unused.
+var (
+	// ff02::1
+	ipv6LinkLocalAllNodes = IP{v6Addr{0: 0xff, 1: 0x02, 15: 0x01}}
+	// ::
+	ipv6Unspecified = IP{v6Addr{}}
+)
+
+// IPv6LinkLocalAllNodes returns the IPv6 link-local all nodes multicast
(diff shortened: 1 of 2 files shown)


bradfitz:
It'd be ideal if we could do this without init work. And something the linker could discard.

mdlayher:
Yeah. I am curious if there is something more clever that could be done, but I also know that being able to check equality via `==` was an explicit goal with this type.

Here's a thought that comes to mind:

```go
type WellKnown int

const (
    Unspecified WellKnown = iota
    IPv6LinkLocalAllNodes
)

func (ip IP) Is(known WellKnown) bool {}

// Usage:
ip, _ := netaddr.ParseIP("::")
fmt.Println(ip.Is(netaddr.Unspecified))
// true
```

bradfitz:
No, I was thinking just laying out the value in memory to help the linker (see https://github.com/golang/go/issues/19533 and https://github.com/golang/go/issues/38450#issuecomment-614248472). Not a new type or API.

Something like:

```go
var ip6Unspecified = IP{v6Addr{}}
func IPv6Unspecified() IP { return ip6Unspecified }
```

Then:

* linker can discard it all (confirm that the unexported var can be discard to and doesn't make an init func?)
* it doesn't allocate at runtuime
* no new types
* people can't modify exported package vars

mdlayher:
I see! I'm good with that.

mdlayher:
Updated.

mdlayher:
Any more thoughts? Thanks!

mdlayher:
I'll take that as my "+2", but feel free to chime in or open issues/PRs if something's amiss! Thanks.

PR #33, opened 2020-05-07
Avoid net.ResolveIPAddr in ParseIP

### LINKED ISSUE #32: ParseIP shouldn't resolve a hostname
The docs don't mention that it will resolve hostnames into IPs, IMO it shouldn't be doing that. I can see the appeal for resolving, but that seems to be better of in a different function.

mmlb:
ParseIP isn't documented to do dns resolution, it really shouldn't
anyway. A resolving function should be named something else, maybe even
ResolveIPAddr to match stdlib.

I originally was falling back to a combination of ParseIP for the ip and
ResolveIPAddr for the zone info, but according to RFC4007 splitting on
'%' is enough.

Signed-off-by: Manuel Mendez <mmendez534@gmail.com>

Fixes #32

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,3 @@
 module inet.af/netaddr
 
 go 1.12
-
-require golang.org/x/perf v0.0.0-20200318175901-9c9101da8316 // indirect
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module inet.af/netaddr
 
 go 1.12
+
+require golang.org/x/perf v0.0.0-20200318175901-9c9101da8316 // indirect
--- a/netaddr.go
+++ b/netaddr.go
@@ -122,13 +122,24 @@ func IPv4(a, b, c, d uint8) IP {
 // s can be in dotted decimal ("192.0.2.1"), IPv6 ("2001:db8::68"),
 // or IPv6 with a scoped addressing zone ("fe80::1cc0:3e8c:119f:c2e1%ens18").
 func ParseIP(s string) (IP, error) {
-	// TODO: do our own parsing to save some allocs? For now,
-	// while showing off new API & representation, just use the
-	// standard library's parsing.
-	ipa, err := net.ResolveIPAddr("ip", s)
-	if err != nil {
-		return IP{}, err
+	var ipa net.IPAddr
+	ipa.IP = net.ParseIP(s)
+	if ipa.IP == nil {
+		// net.ParseIP can't deal with zoned scopes, lets split and try to parse the ip again
+		if index := strings.Index(s, "%"); index > -1 {
+			// handle bad string with % at the end
+			if index == len(s)-1 {
+				return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
+			}
+			ipa.Zone = s[index+1:]
+			s = s[:index]
+		}
+		ipa.IP = net.ParseIP(s)
+		if ipa.IP == nil {
+			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse ip", s)
+		}
 	}
+
 	if !strings.Contains(s, ":") {
 		if ip4 := ipa.IP.To4(); ip4 != nil {
 			var v4 v4Addr
--- a/netaddr.go
+++ b/netaddr.go
@@ -125,18 +125,24 @@ func ParseIP(s string) (IP, error) {
 	var ipa net.IPAddr
 	ipa.IP = net.ParseIP(s)
 	if ipa.IP == nil {
-		// net.ParseIP can't deal with zoned scopes, lets split and try to parse the ip again
-		if index := strings.Index(s, "%"); index > -1 {
-			// handle bad string with % at the end
-			if index == len(s)-1 {
-				return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
+		switch percent := strings.Index(s, "%"); percent {
+		case -1:
+			// handle bad input with no % at all, so the net.ParseIP was not due to a zoned IPv6 fail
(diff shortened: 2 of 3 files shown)


mmlb:
Allocs went down too:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkParseIPv4-8        255           75.5          -70.39%
BenchmarkParseIPv6-8        332           211           -36.45%
BenchmarkStdParseIPv4-8     44.1          44.0          -0.23%
BenchmarkStdParseIPv6-8     75.4          75.8          +0.53%

benchmark                   old allocs     new allocs     delta
BenchmarkParseIPv4-8        5              2              -60.00%
BenchmarkParseIPv6-8        5              3              -40.00%
BenchmarkStdParseIPv4-8     1              1              +0.00%
BenchmarkStdParseIPv6-8     1              1              +0.00%

benchmark                   old bytes     new bytes     delta
BenchmarkParseIPv4-8        132           20            -84.85%
BenchmarkParseIPv6-8        160           64            -60.00%
BenchmarkStdParseIPv4-8     16            16            +0.00%
BenchmarkStdParseIPv6-8     16            16            +0.00%
```

mmlb:
@bradfitz fyi

bradfitz:
Thanks! Will look soon.

netaddr.go
```diff
@@ -122,13 +122,24 @@ func IPv4(a, b, c, d uint8) IP {
...
-	// while showing off new API & representation, just use the
-	// standard library's parsing.
-	ipa, err := net.ResolveIPAddr("ip", s)
-	if err != nil {
-		return IP{}, err
+	var ipa net.IPAddr
+	ipa.IP = net.ParseIP(s)
+	if ipa.IP == nil {
+		// net.ParseIP can't deal with zoned scopes, lets split and try to parse the ip again
+		if index := strings.Index(s, "%"); index > -1 {
```
bradfitz, line 129: I'd name this variable either `i` (which is the conventional `index` name), or something more meaningful like `percent`.

Also, `!= -1` is more common.

```go
if percent := strings.Index(s, "%"); percent != -1 {
      s, ipa.Zone = s[:percent], s[percent+1:]
      if ipa.Zone == "" {
           return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
      }
      ipa.IP = net.ParseIP(s)
}
if ipa.IP == nil {
	return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse IP", s)
}
```

netaddr.go
```diff
@@ -122,13 +122,24 @@ func IPv4(a, b, c, d uint8) IP {
...
+			// handle bad string with % at the end
+			if index == len(s)-1 {
+				return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
+			}
+			ipa.Zone = s[index+1:]
+			s = s[:index]
+		}
+		ipa.IP = net.ParseIP(s)
+		if ipa.IP == nil {
+			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse ip", s)
```
bradfitz, line 139: capitalize IP

netaddr_test.go
```diff
@@ -738,10 +738,6 @@ func TestParseIPPrefixError(t *testing.T) {
 			prefix: "192.168.0.0",
 			errstr: "no '/'",
 		},
-		{
```
bradfitz, line 741: please don't delete test cases. You can update them to show the new behavior, though.

netaddr.go
mmlb: fixed

netaddr_test.go
mmlb: fixed the err string, but this seems to be testing lower than what ParseIPPrefix should be concerned with right? This test should (and is) be a case for ParseIP.

mmlb:
Did the fixes as a fixup commit instead of fixing directly to ease review, let me know if I should squash instead.

mmlb:
oops my Benchmarks came in too by mistake in the fixup. I guess that brings up the question, should I drop them or keep them (and maybe relocate them in the file somewhere else for grouping's sake?)

go.mod
```diff
@@ -1,3 +1,5 @@
 module inet.af/netaddr
 
 go 1.12
+
+require golang.org/x/perf v0.0.0-20200318175901-9c9101da8316 // indirect
```
bradfitz, line 5: Remove this. `go mod tidy`.

netaddr.go
```diff
@@ -122,13 +122,30 @@ func IPv4(a, b, c, d uint8) IP {
...
+			// handle bad input with no % at all, so the net.ParseIP was not due to a zoned IPv6 fail
+			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse IP", s)
+		case 0:
+			// handle bad input with % at the start
+			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing IPv6 address", s)
+		case len(s) - 1:
+			// handle bad input with % at the end
+			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
+		default:
+			// net.ParseIP can't deal with zoned scopes, lets split and try to parse the ip again
```
bradfitz, line 139: typo: let's

and IP capitalized

netaddr.go
```diff
@@ -122,13 +122,30 @@ func IPv4(a, b, c, d uint8) IP {
...
+			ipa.Zone = s[percent+1:]
```
bradfitz, line 140: put these two lines on the same line like I wrote before

```
s, ipa.zone = ...
```

go.mod
mmlb: done

netaddr.go
mmlb: fixed

netaddr.go
mmlb: didn't catch that last go round, fixed.

PR #45, opened 2020-07-18
Don't panic on zero-value IP in Contains

dshynkev:
Another reasonable behavior is to return `p.IP.ipImpl == nil`, considering a zero-value IP to be included in a zero-value Prefix.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -727,8 +727,12 @@ func (p IPPrefix) IPNet() *net.IPNet {
 //
 // An IPv4 address will not match an IPv6 prefix.
 // A 4-in-6 IP will not match an IPv4 prefix.
+// A zero-value IP will not match any prefix.
 func (p IPPrefix) Contains(addr IP) bool {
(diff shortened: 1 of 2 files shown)


netaddr.go
```diff
@@ -727,8 +727,12 @@ func (p IPPrefix) IPNet() *net.IPNet {
 //
 // An IPv4 address will not match an IPv6 prefix.
 // A 4-in-6 IP will not match an IPv4 prefix.
+// A zero-value IP will not match any prefix.
```
bradfitz, line 730: Actually, I already merged this, but reading this doc again, it could be read as meaning p.IP can also be zero to make this return false.

Currently a zero p.IP will still panic.

dshynkev: I thought it was clear from the IP <-> prefix separation also found in lines above ("X IP will not match Y prefix" <-> "X `addr` will not match Y `p`"). But maybe a zero `p.IP` should return false for all IPs too?

PR #46, opened 2020-08-09
zero masked address bits returned from ParseIPPrefix

eliasnaur:
Fixes tailscale/tailscale#645

Signed-off-by: Elias Naur <mail@eliasnaur.com>

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -681,6 +681,8 @@ func FromStdIPNet(std *net.IPNet) (prefix IPPrefix, ok bool) {
 // ParseIPPrefix parses s as an IP address prefix.
 // The string can be in the form "192.168.1.0/24" or "2001::db8::/32",
 // the CIDR notation defined in RFC 4632 and RFC 4291.
+//
+// ParseIPPrefix zeroes masked address bits.
 func ParseIPPrefix(s string) (IPPrefix, error) {
 	i := strings.IndexByte(s, '/')
 	if i < 0 {
@@ -702,10 +704,7 @@ func ParseIPPrefix(s string) (IPPrefix, error) {
 	if bits < 0 || bits > maxBits {
 		return IPPrefix{}, fmt.Errorf("netaddr.ParseIPPrefix(%q): prefix length out of range", s)
 	}
-	return IPPrefix{
-		IP:   ip,
-		Bits: uint8(bits),
-	}, nil
+	return ip.Prefix(uint8(bits))
 }
 
 // IPNet returns the net.IPNet representation of an IPPrefix.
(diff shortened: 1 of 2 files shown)


bradfitz:
I'd like @mdlayher and @danderson's opinions on this before merging.

It should probably be documented whichever direction we go, though.

mdlayher:
A counter-point I just thought of: the type is IPPrefix, so the "prefix" part is explicitly noted. Tough call. I'm still slightly leaning toward keeping it as is though.

eliasnaur:
> I'm -1 on this one: it's easy for the caller to invoke Prefix themselves to retrieve the masked output.
> 
> I have a similar complaint with net.ParseCIDR in the stdlib since it does something similar by splitting apart the individual IP address and its IPNet network, which has resulted in actual subtle bugs in my/other's programs.
> 

I don't have a strong opinion, but I was surprised that address.Prefix zeroes masked bits but ParseIPPrefix doesn't. In other
words, ParseIPPrefix sounds to me as a shorthand for ParseIP on the address part, followed by Prefix applied on the mask.

> Either way we go, we should explicitly document the expectations.

Added a note about zeroing.

mdlayher:
I took a quick look at a couple of places where I'm using this function. I have at least one case where I want to return an error if the caller does not pass a valid prefix, and another which has a bug because I need a valdi prefix but am not validating by calling the Prefix method.

So it seems to me that mirroring the stdlib with `func ParseIPPrefix(s string) (IP, IPPrefix, error)` may be the right call after all as it supports both use cases without any loss of information.

bradfitz:
Seems like a common way to write your IP & netmask at the same time is to, say, `10.0.0.52/24`. Currently we can parse that and keep both bits of information. With this change we can't and the user has to do their own splitting on `/`. Maybe it doesn't matter too much.

I think I'm slightly in favor of not changing behavior (usually a nice feature) and instead documenting this.

We could add a method on `IPPrefix` like `func (p IPPrefix) Masked() IPPrefix` that returns `p` with `p.IP` masked, like:

```go
// Masked returns p in its canonical form, with bits of of p.IP not in p.Bits masked off.
// If p is zero or otherwise invalid, Masked returns the zero value.
func (p IPPrefix) Masked() IPPrefix {
     if m, err := p.IP.Prefix(p.Bits); err == nil {
        return m
     }
     return IPPrefix{}
}
```

mdlayher:
A convenience function (and documentation to point users toward it) LGTM.

eliasnaur:
Here you go: https://github.com/inetaf/netaddr/pull/47.

mdlayher:
Nice work, thanks! Closing as this is superseded by #47.

PR #48, opened 2020-10-30
Add method IPPrefix.Overlaps(IPPrefix) bool

bradfitz:


bradfitz:
cc @sailorfrag

netaddr_test.go
```diff
@@ -1323,3 +1323,33 @@ func TestAs4(t *testing.T) {
...
 	}
 }
+
+func TestIPPrefixOverlaps(t *testing.T) {
+	pfx := mustIPPrefix
+	tests := []struct {
+		a, b IPPrefix
+		want bool
+	}{
+		{IPPrefix{}, pfx("1.2.0.0/16"), false},  // first zero
```
sailorfrag, line 1333: It seems to me like you'd want 0/0 (or 0.0.0.0/0 if it doesn't allow omitting the unnecessary IP parts) to return true for anything, e.g. in cases where you're checking if something matches a default route.

netaddr.go
```diff
@@ -778,6 +780,45 @@ func (p IPPrefix) Contains(addr IP) bool {
...
+//
+// If either has a Bits of zero, it returns true.
+func (p IPPrefix) Overlaps(o IPPrefix) bool {
+	if p.IP.IsZero() || o.IP.IsZero() {
+		return false
+	}
+	if p == o {
+		return true
+	}
+	if p.IP.Is4() != o.IP.Is4() {
```
sailorfrag, line 796: It might be worth a comment mentioning what you intend to happen here with IPv6-mapped IPv4 addresses.

netaddr_test.go
```diff
@@ -1323,3 +1323,33 @@ func TestAs4(t *testing.T) {
...
+		{pfx("1.2.3.0/24"), pfx("1.2.0.0/16"), true},
+
+		{pfx("1.2.0.0/16"), pfx("1.2.3.0/32"), true},
+		{pfx("1.2.3.0/32"), pfx("1.2.0.0/16"), true},
+
+		{pfx("1.2.3.0/32"), pfx("5.5.5.5/0"), true}, // normalization not required; /0 means true
+
+		{pfx("5::1/128"), pfx("5::0/8"), true},
+		{pfx("5::0/8"), pfx("5::1/128"), true},
+	}
```
sailorfrag, line 1349: Add a negative IPv6/IPv6 test?

netaddr_test.go
bradfitz: Note that this is a zero value, not a value representing a `/0`.

The contract as documented is you must provide two valid values to compare. Otherwise we just return false. (The alternative perhaps might be to panic, but that seems aggressive.)

netaddr.go
bradfitz: Right! Will do.

netaddr_test.go
sailorfrag: Oh, if 0.0.0.0/0 and IsZero() are distinct, then that's great. Maybe just add a test verifying that it does indeed work, since that seems like the sort of subtle thing that could break by accident.

bradfitz:
PTAL

netaddr.go
```diff
@@ -778,6 +780,46 @@ func (p IPPrefix) Contains(addr IP) bool {
 	return true
 }
 
+// Overlaps reports whether p and o overlap at all.
+//
+// If p and o are of different address families or either have a zero
+// IP, it reports false. Like the Contains method, a prefix with a
+// 4-in-6 IP is still treated as an IPv6 mask.
```
danderson, line 787: 4-in-6 is an IPv6 transitional mechanism, not the "embed v4 in v6" thing. I think you want to say "a prefix with an IPv6-mapped IPv4 IP is ..."

netaddr_test.go
```diff
@@ -1323,3 +1323,45 @@ func TestAs4(t *testing.T) {
...
+
+		// IPv6 overlapping
+		{pfx("5::1/128"), pfx("5::0/8"), true},
+		{pfx("5::0/8"), pfx("5::1/128"), true},
+
+		// IPv6 not overlapping
+		{pfx("1::1/128"), pfx("2::2/128"), false},
+		{pfx("0100::0/8"), pfx("::1/128"), false},
+
+		// IPv4-in-6 should not overlap with IPv4.
```
danderson, line 1359: v6-mapped v4

danderson (approved):

mdlayher (commented):
LGTM!

PR #49, opened 2020-11-22
Add IPRange, IPRangeSet & friends to work on sets of prefixes/ranges

bradfitz:
Not done, but you can see where I'm going. Just need to map ranges to a covering of `[]IPPrefix` now.

/cc @josharian @danderson @mdlayher @crawshaw

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -11,7 +11,9 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"log"
 	"net"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -45,7 +47,7 @@ type ipImpl interface {
 	is4() bool
 	is6() bool
 	is4in6() bool
-	as16() [16]byte
+	as16() ip16
 	// prefix is the type-specific implementation of IP.Prefix.
 	prefix(uint8) (IPPrefix, error)
 	String() string
@@ -56,8 +58,8 @@ type v4Addr [4]byte
 func (v4Addr) is4() bool    { return true }
 func (v4Addr) is6() bool    { return false }
 func (v4Addr) is4in6() bool { return false }
-func (ip v4Addr) as16() [16]byte {
-	return [16]byte{
+func (ip v4Addr) as16() ip16 {
+	return ip16{
 		10: 0xff,
 		11: 0xff,
 		12: ip[0],
@@ -92,10 +94,10 @@ const (
 
 type v6Addr [16]byte
 
-func (v6Addr) is4() bool         { return false }
-func (v6Addr) is6() bool         { return true }
-func (ip v6Addr) is4in6() bool   { return string(ip[:len(mapped4Prefix)]) == mapped4Prefix }
-func (ip v6Addr) as16() [16]byte { return ip }
+func (v6Addr) is4() bool       { return false }
+func (v6Addr) is6() bool       { return true }
+func (ip v6Addr) is4in6() bool { return string(ip[:len(mapped4Prefix)]) == mapped4Prefix }
+func (ip v6Addr) as16() ip16   { return ip16(ip) }
 func (ip v6Addr) prefix(bits uint8) (IPPrefix, error) {
 	if bits > 128 {
 		return IPPrefix{}, fmt.Errorf("netaddr: prefix length %d too large for IP address family", bits)
@@ -288,35 +290,47 @@ func (ip IP) Zone() string {
 	return ""
 }
 
-// Less reports whether ip sorts before ip2.
-// IP addresses sort first by length, then their address.
-// IPv6 addresses with zones sort just after the same address without a zone.
-func (ip IP) Less(ip2 IP) bool {
+// Compare returns an integer comparing two IPs.
+// The result will be 0 if ip==ip2, -1 if ip < ip2, and +1 if ip > ip2.
+// The definition of "less than" is the same as the IP.Less method.
+func (ip IP) Compare(ip2 IP) int {
 	a, b := ip, ip2
 	// Zero value sorts first.
 	if a.ipImpl == nil {
-		return b.ipImpl != nil
+		if b.ipImpl == nil {
+			return 0
+		}
+		return -1
 	}
 	if b.ipImpl == nil {
-		return false
+		return 1
 	}
 
 	a4, b4 := a.Is4(), b.Is4()
 	if a4 != b4 {
-		// When address families differ, IPv4 sorts first.
-		return a4
+		if a4 {
+			return -1
+		}
+		return 1
 	}
 
 	aa, ba := a.As16(), b.As16()
-	switch bytes.Compare(aa[:], ba[:]) {
-	case -1:
-		return true
-	case 1:
-		return false
-	default: // case 0 (bytes.Compare only returns -1, 1, or 0)
-		return !a4 && a.Zone() < b.Zone()
+	c := bytes.Compare(aa[:], ba[:])
+	if c == 0 && !a4 {
+		za, zb := a.Zone(), b.Zone()
+		if za < zb {
+			c = -1
+		} else if za > zb {
+			c = 1
+		}
 	}
+	return c
 }
+
+// Less reports whether ip sorts before ip2.
+// IP addresses sort first by length, then their address.
+// IPv6 addresses with zones sort just after the same address without a zone.
+func (ip IP) Less(ip2 IP) bool { return ip.Compare(ip2) == -1 }
 
 // ipZone returns the standard library net.IP from ip, as well
 // as the zone.
@@ -715,6 +729,17 @@ func (p IPPrefix) Masked() IPPrefix {
 	return IPPrefix{}
 }
 
+// Range returns the inclusive range of IPs that p covers.
+//
+// If p is zero or otherwise invalid, Range returns the zero value.
+func (p IPPrefix) Range() IPRange {
+	p = p.Masked()
+	if p.IsZero() {
+		return IPRange{}
+	}
+	return IPRange{From: p.IP, To: p.LastIP()}
+}
+
 // IPNet returns the net.IPNet representation of an IPPrefix.
 // The returned value is always non-nil.
 // Any zone identifier is dropped in the conversion.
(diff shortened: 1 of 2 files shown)


josharian:
Is this intended to replace the Cilium manipulation code I started hacking on on Friday?

bradfitz:
It could. I needed a fun little exercise last night and had been thinking about it.

bradfitz:
Okay, PTAL. The hard bits are all done now and the main function we need (albeit with a kinda ugly signature) is now there with tests. There are some API decisions to make, though:

* export `IPRange`? ... useful on its own probably?
* make a new IPRangeSet type that's opaque with methods like:
   - `func (s *IPRangeSet) AddPrefix(p IPPrefix)`
   - `func (s *IPRangeSet) AddRange(r IPRange)`
   - `func (s *IPRangeSet) Add(b IPRangeSet)`
   - `func (s *IPRangeSet) Sub(b IPRangeSet)`
   - `func (s *IPRangeSet) Prefixes() []IPPrefix`
   - `func (s *IPRangeSet) Ranges() []IPRange`
   - `func (s *IPRangeSet) Matcher() (contains func(IP) bool)` (could generate an optimized matcher for re-use, like `strings.NewReplacer`, `regexp.Compile`, etc)

If that API sounds good, doing it is easy now that all the necessary code is in this draft PR at this point (just unexported).

I could also unify some of the various code internally (use `ip16` type for more things, unify the `IPPrefix.Masked` impl, etc)

netaddr.go
```diff
@@ -846,3 +848,297 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+// prefixes with all the 'remove' prefixes removed.
+func MergeRemovePrefixes(add, remove []IPPrefix) []IPPrefix {
+	var out []IPPrefix
+	for _, ipr := range mergeRemoveRanges(add, remove) {
+		out = append(out, ipr.Prefixes()...)
+	}
+	return out
+}
+
+// TODO: export this?
```
crawshaw, line 886: SGTM

crawshaw (commented):
Started reading through the logic, so far looks good though it would be nice if there were some more robust way to measure the effects. Maybe a random walk through the API you're proposing.

Re: APIs: the one you propose is much easier to think about than the one currently in the PR, so +1 to exporting IPRange.

netaddr.go
```diff
@@ -846,3 +848,297 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
 func (p IPPrefix) String() string {
 	return fmt.Sprintf("%s/%d", p.IP, p.Bits)
 }
+
+// LastIP returns the last IP in the prefix.
```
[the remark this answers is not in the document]
bradfitz: done

netaddr.go
```diff
@@ -288,6 +289,42 @@ func (ip IP) Zone() string {
 	return ""
 }
 
+// Compare returns an integer comparing two IPs. The result will be 0 if ip==ip2, -1 if ip < ip2, and +1 if ip > ip2.
+// The definition of "less than" is the same as the IP.Less method.
```
crawshaw, line 293: Any reason not to replace the implementation of Less with `return ip.Compare(ip2) == -1`?

netaddr.go
```diff
@@ -846,3 +894,374 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+func (ip *ip16) clear(bit uint8) {
+	i, s := bit/8, 7-(bit%8)
+	ip[i] &^= 1 << s
+}
+
+// lastWithBitZero returns a copy of ip with the given bit
+// cleared and the following all set.
+func (ip ip16) lastWithBitZero(bit uint8) ip16 {
+	ip.clear(bit)
+	for ; bit < 128; bit++ {
```
crawshaw, line 971: The potential optimization here made me go searching for a Go bit vector package, but amazingly, there doesn't seem to be a good one? Hard to believe.

josharian: Using encoding/binary and then math/bits should generate pretty good code.

josharian (commented):
I look forward to reviewing later today. Quick response in the meantime...

netaddr.go
bradfitz: done.

netaddr.go
bradfitz: done

netaddr.go
bradfitz: Can optimize later.

bradfitz:
Ready for review.

crawshaw (approved):
Very nice!

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
 func (p IPPrefix) String() string {
 	return fmt.Sprintf("%s/%d", p.IP, p.Bits)
 }
+
+// LastIP returns the last IP in the prefix.
+func (p IPPrefix) LastIP() IP {
```
danderson, line 876: Bit weird to have this exported by itself. Maybe pair with FirstIP if you want this to be a public API? Or unexport.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+	// It must have the same address family (IPv4 vs IPv6) as From
+	// and be greater than or equal to From.
+	// To is an inclusive bound; it is included in the set.
+	To IP
+}
+
+// Valid reports whether r.From and r.To are both non-zero and obey
+// the documented requirements: address families match, and From is
+// less than or equal to To.
+func (r IPRange) Valid() bool {
```
mdlayher, line 917: Good call, I was going to ask why From/To were exported if the less-than invariant must hold. Can you think of any use case in particular where someone would construct an IPRange manually without a constructor function? If not, it seems tempting to me to unexport the fields.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
 func (p IPPrefix) String() string {
 	return fmt.Sprintf("%s/%d", p.IP, p.Bits)
 }
+
+// LastIP returns the last IP in the prefix.
+func (p IPPrefix) LastIP() IP {
+	if p.IP.IsZero() {
+		return IP{}
+	}
+	a16 := p.IP.As16()
```
danderson, line 880: I have some bit twiddling math in wgengine/filter that would work a treat on these functions. Will send a followup PR to port things over.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+	} else {
+		return IPv6Raw(a16) // doesn't unmap
+	}
+}
+
+// IPRange represents an inclusive range of IP addresses
+// from the same address family.
+type IPRange struct {
+	// From is the starting IP address in the range.
+	// It must have the same address family (IPv4 vs IPv6) as To
```
danderson, line 902: What happens if the contract is violated? AFAICT from below, we return a zero-ish value rather than, say, panic. Would be nice to document here.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+	// and be less than or equal to To.
+	// From is an inclusive bound; it is included in the set.
```
danderson, line 904: included in the range, not set. Same for To's docstring.

mdlayher (commented):
LGTM. Definitely some complex code going in here, but that's to be expected with this sort of low-level bit manipulation. Did you run go-fuzz on this by chance?

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+	} else {
+		ok = subOne(a[:], 15)
+	}
+	if ok {
+		return ipFrom16Match(ip, a)
+	}
+	return IP{}
+}
+
+// IPRangeSet represents a set of IPRanges.
```
danderson, line 1086: Could call this just IPSet, since those are the semantics being implemented. The fact that it's IPRanges on the inside is irrelevant.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+		}
+		if p.start {
+			log.Printf(" {  %-15s %s\n", p.ip, emo)
+		} else {
+			log.Printf("  } %-15s %s\n", p.ip, emo)
+		}
+	}
+}
+
+// Ranges removes the minimum and sorted set of IP
```
danderson, line 1165: returns, not removes.

danderson (commented):
Some API thoughts, mostly. Could be made much faster, but correct >> fast. Nice!

netaddr.go
bradfitz: Oh, I meant to unexport this once I added the Range method.

netaddr.go
bradfitz: updated docs a bit.

netaddr.go
bradfitz: done.

netaddr.go
bradfitz: Well, we have IPPort already that's also with two exported fields. Felt consistent?

netaddr.go
bradfitz: done.

netaddr.go
bradfitz: Also IPPrefix, which can have invalid values (bits over 128) from literal construction.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+// set and the following all cleared.
+func (ip ip16) firstWithBitOne(bit uint8) ip16 {
+	ip.set(bit)
+	for ; bit < 128; bit++ {
+		ip.clear(bit)
+	}
+	return ip
+}
+
+// prefixMaker returns a address-family-corrected IPPrefix from ip16 and bits,
```
josharian, line 964: an

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+}
+
+func appendRangePrefixes(dst []IPPrefix, makePrefix prefixMaker, a16, b16 ip16) []IPPrefix {
+	common := uint8(0)
+	for common < 128 && a16.bitSet(common) == b16.bitSet(common) {
+		common++
+	}
+	// See whether a16 and b16, after their common shared bits, end
+	// in all zero bits or all one bits, respectively.
+	aAllZero, bAllSet := true, true
```
josharian, line 997: This can be a single bool; you can combine the following two loops.

```go
allZeroOne := true
for i := common; i < 128; i++ {
  if a16.bitSet(i) || !b16.bitSet(i) {
    allZeroOne = false
    break
  }
}
```

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+		// (a16 being 10.50.0.0 and b16 being 10.50.255.255)
+		return append(dst, makePrefix(a16, common))
+	}
+	// Otherwise recursively do both halves.
+	dst = appendRangePrefixes(dst, makePrefix, a16, a16.lastWithBitZero(common+1))
+	dst = appendRangePrefixes(dst, makePrefix, b16.firstWithBitOne(common+1), b16)
+	return dst
+}
+
+func addOne(a []byte, i int) bool {
```
josharian, line 1021: This is fine as-is, but an alternative is to accept just a byte slice and use the end of the slice to locate i. Then reslice to recurse (or better make iterative). Then call sites look like `addOne(a[12:15])` instead of `addOne(a[12:], 3)`, which strikes me as a bit clearer.

josharian: It's a pity math/big.nat isn't exported, 'cause it'd be perfect for a lot of these operations. Inc, AndNot, etc.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+		return ipFrom16Match(ip, a)
+	}
+	return IP{}
+}
+
+// IPRangeSet represents a set of IPRanges.
+//
+// The zero value is a valid value representing a set of no IP ranges.
+//
+// The Add and Remove methods add or remove ranges to the set.  Add
```
josharian, line 1090: remove double space after period

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+// methods should be called first, as a remove operation does nothing
```
josharian, line 1091: This makes me a bit nervous. It seems too easy to misuse as an API. Maybe make the invariant stronger: Calling Remove after calling Add causes a panic. Or put another way: You must add all desired ranges, then remove any undesired ranges.

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+	})
+	const debug = false
+	if debug {
+		log.Printf("post-sort:")
+		debugLogPoints(points)
+		log.Printf("merging...")
+	}
+
+	// Now build 'want', like points but with "remove" ranges removed
+	// and adjancent blocks merged, and all elements alternating between
```
josharian, line 1198: adjacent

netaddr.go
```diff
@@ -846,3 +871,422 @@ func (p *IPPrefix) UnmarshalText(text []byte) error {
...
+			From: want[i].ip,
+			To:   want[i+1].ip,
+		})
+	}
+	return out
+}
+
+// Prefixes returns the minimum and sorted set of IP prefixes
+// that covers s.
+// returning a new slice of prefixes that covers all of the given 'add'
```
josharian, line 1284: delete two stray lines?

josharian (commented):
Nice.

PR #50, opened 2020-11-23
Add IPRangeSet.ContainsFunc

bradfitz:
cc @josharian too

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -1290,3 +1290,28 @@ func (s *IPRangeSet) Prefixes() []IPPrefix {
 	}
 	return out
 }
+
+// ContainsFunc returns a func that reports whether an IP is in s.
+// The returned func opertes on a copy of s, so s may be mutated
+// later.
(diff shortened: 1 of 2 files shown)


crawshaw (approved):

netaddr.go
```diff
@@ -1290,3 +1290,28 @@ func (s *IPRangeSet) Prefixes() []IPPrefix {
 	}
 	return out
 }
+
+// ContainsFunc returns a func that reports whether an IP is in s.
+// The returned func opertes on a copy of s, so s may be mutated
```
mdlayher, line 1295: *operates

And why make a copy? For fine-grained allocation control, shouldn't it be up to the caller to decide whether or not to make a copy?

bradfitz: Fixed typo, thanks!

The idea is that the `Add*` and `Remove*` methods can be cheap and we only need to compile it into a optimized machine once at the end.

So the point isn't so much the copy, but that the machine is already saved so mutations to s won't affect it.

PR #55, opened 2020-12-17
WIP: change representation to 24 bytes

### LINKED ISSUE #53: Change representation of IP?
I think there's still some general unhappiness that our IP type requires an allocation to box the `[4]byte` or `[16]byte` into an interface.

I've previously pushed back against any change to the `IP` type's representation because we have 4 conflicting requirements that I never saw the way out of:

1. must be comparable
2. must support IPv6 zones
3. must be able to distinguish IPv4 and IPv4-mapped IPv6 addresses (that is, `1.2.3.4` and `::ffff::1.2.3.4` should be different)
4. shouldn't be bigger than a slice (24 bytes: three 8-byte pointers), which is my upper bound for tolerable value types

We also discussed the solution to requirement (3) as needing "a bool", but because an IPv4 address can't have a zone, we can re-use a sentinel zone value as meaning "actually an IPv4 address".

Which means we have:

```go
    type IP struct {
        addr [16]byte
        zone ZT
    }
```

The question then is: what type is the zone, `ZT`? If the goal is 24 bytes, we only have 8 bytes to work with, so the natural choice of `string` (at 16 bytes: a pointer and int) is out. That would bump it up to 32 bytes. (If we were cool with 32-byte IP addresses, that's an answer, but I'm not particularly happy about that.)

But if we used an integer for the zone, we could be smaller, though.

On Windows, a network adapter's local ID (LUID) is a ULONG64 (uint64): https://docs.microsoft.com/en-us/windows/win32/api/ifdef/ns-ifdef-net_luid_lh

Maybe that's sufficient? We'd need to find a sentinel zone uint64 number to use for "actually an IPv4 address".

It does mean that we wouldn't be able to map an arbitrary https://golang.org/pkg/net/#IPAddr Zone string,

```go
type IPAddr struct {
    IP   IP
    Zone string // IPv6 scoped addressing zone; added in Go 1.1
}
```

... into our `IP` type if the `Zone` was not the name of a local interface. Maybe that's okay?

The Go standard library maintains a cache mapping between integers and strings internally:

https://github.com/golang/go/blob/go1.16beta1/src/net/interface.go#L178

We could too.

Thoughts, @crawshaw @danderson @mdlayher?

bradfitz:
Fixes #53

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -29,50 +29,40 @@ const (
 	maxUint16 = 1<<16 - 1
 )
 
+// o4 is the offset to the beginning of the IPv4 address within IP.a.
+const o4 = 12
+
 // Sizes: (64-bit)
 //   net.IP:     24 byte slice header + {4, 16} = 28 to 40 bytes
 //   net.IPAddr: 40 byte slice header + {4, 16} = 44 to 56 bytes + zone length
-//   netaddr.IP: 16 byte interface + {4, 16, 24} = 20, 32, 40 bytes + zone length
+//   netaddr.IP: 24 bytes, always (zone, if any, is cached globally forever)
 
 // IP represents an IPv4 or IPv6 address (with or without a scoped
 // addressing zone), similar to Go's net.IP or net.IPAddr.
 //
 // Unlike net.IP or net.IPAddr, the netaddr.IP is a comparable value
 // type (it supports == and can be a map key) and is immutable.
-// Its memory representation ranges from 20 to 40 bytes, depending on
-// whether the underlying adddress is IPv4, IPv6, or IPv6 with a
-// zone. (This is smaller than the standard library's 28 to 56 bytes)
+// Its memory representation is 24 bytes in 64-bit machines (the same
+// size as a Go slice header) for both IPv4 and IPv6 address.
 type IP struct {
-	ipImpl
+	a  [16]byte // IPv6 or IPv4-mapped IPv6 form of IPv4 addr
+	zt ztype    // 0=invalid, 1=IPv4, 2=zoneless IPv6, 3+ for IPv6 w/ zone from table
 }
 
-// ipImpl is the interface representing either a v4addr, v6addr, v6ZoneAddr.
-type ipImpl interface {
-	is4() bool
-	is6() bool
-	is4in6() bool
-	as16() ip16
-	// prefix is the type-specific implementation of IP.Prefix.
-	prefix(uint8) (IPPrefix, error)
-	String() string
-}
+// ztype is the zone index or type.
+type ztype uint64
 
-type v4Addr [4]byte
+const (
+	z0    = ztype(0)
+	z4    = ztype(1)
+	z6noz = ztype(2) // no zone
+)
 
-func (v4Addr) is4() bool    { return true }
-func (v4Addr) is6() bool    { return false }
-func (v4Addr) is4in6() bool { return false }
-func (ip v4Addr) as16() ip16 {
-	return ip16{
-		10: 0xff,
-		11: 0xff,
-		12: ip[0],
-		13: ip[1],
-		14: ip[2],
-		15: ip[3],
-	}
-}
-func (ip v4Addr) prefix(bits uint8) (IPPrefix, error) {
+func (z ztype) IsZero() bool { return z == z0 }
+func (z ztype) Is4() bool    { return z == z4 }
+func (z ztype) Is6() bool    { return z > z4 }
+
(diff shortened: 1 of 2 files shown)


bradfitz:
WIP; checkpoint for a few hours.

/cc @danderson

netaddr.go
```diff
@@ -412,96 +361,76 @@ func (ip IP) Is6() bool {
...
+	if ip.zt == z0 {
 		return IPPrefix{}, nil
 	}
-	// TODO: optimize this to return ip directly if it's already
-	// masked and to not allocate a new IP.ipImpl.
-	return ip.prefix(bits)
+	if ip.Is4() {
+		return v4Prefix(ip.As4(), bits)
+	}
+	// TODO: preserve zone?
```
danderson, line 432: we should delegate that decision to @mdlayher, as corerad is the most zone-aware thing we know of using netaddr. Matt, does zone in a prefix make any sense to have?

Barring a known use case, I'm inclined to keep the zone anyway, because it can be used to e.g. mean "the link-local fe80:: prefix on a specific interface", although I can't think how I'd use it for same.

danderson (commented):
Looks pretty neat so far, and surprisingly (to me) not that much to change.

netaddr.go
mdlayher: I can't think of any reason you'd need zone on a prefix. I don't think I've ever seen eg "fe80::%eth0/64" or similar. But I'll think on it.

bradfitz: Actually we had tests that it was preserved (that I failed), so I put it back.

bradfitz:
PTAL. All tests pass now.

danderson:
Got benchstat results? Inquiring minds want to know.

netaddr.go
```diff
@@ -29,50 +29,40 @@ const (
...
 
 // IP represents an IPv4 or IPv6 address (with or without a scoped
 // addressing zone), similar to Go's net.IP or net.IPAddr.
 //
 // Unlike net.IP or net.IPAddr, the netaddr.IP is a comparable value
 // type (it supports == and can be a map key) and is immutable.
-// Its memory representation ranges from 20 to 40 bytes, depending on
-// whether the underlying adddress is IPv4, IPv6, or IPv6 with a
-// zone. (This is smaller than the standard library's 28 to 56 bytes)
+// Its memory representation is 24 bytes in 64-bit machines (the same
```
danderson, line 45: It's going to be 24 bytes on all machines, right? There are no more machine-dependent types in there.

netaddr.go
```diff
@@ -29,50 +29,40 @@ const (
...
-	// prefix is the type-specific implementation of IP.Prefix.
-	prefix(uint8) (IPPrefix, error)
-	String() string
-}
+// ztype is the zone index or type.
+type ztype uint64
 
-type v4Addr [4]byte
+const (
+	z0    = ztype(0)
```
danderson, line 56: Add suffix comments for `// invalid` and `// IPv4` ?

netaddr.go
```diff
@@ -383,27 +317,18 @@ func (ip IP) IPAddr() *net.IPAddr {
...
+	return ip.zt == z4
 }
 
 // Is4in6 reports whether ip is an IPv4-mapped IPv6 address.
 func (ip IP) Is4in6() bool {
-	if ip.ipImpl == nil {
-		return false
-	}
-	return ip.ipImpl.is4in6()
+	return ip.zt == z6noz && string(ip.a[:len(mapped4Prefix)]) == mapped4Prefix
```
danderson, line 325: Should v6+zone return true as well?

danderson (approved):

netaddr.go
```diff
@@ -510,25 +437,21 @@ func (ip IP) Prefix(bits uint8) (IPPrefix, error) {
...
 }
 
 // As4 returns an IPv4 or IPv4-in-IPv6 address in its 4 byte representation.
 // If ip is the IP zero value or an IPv6 address, As4 panics.
 // Note that 0.0.0.0 is not the zero value.
 func (ip IP) As4() [4]byte {
-	if ip.ipImpl == nil {
+	switch ip.zt {
+	case z4:
+		return [4]byte{ip.a[o4+0], ip.a[o4+1], ip.a[o4+2], ip.a[o4+3]}
```
mdlayher, line 449: Seems like we do this offset addition logic a fair amount, could be worth splitting out a method.

netaddr.go
danderson: We dug into this separately, and turns out v4 link-local addresses are supposed to be used by binding to the specific interface you want to send out of, rather than using a zone specifier. Yay incompatible APIs!

PR #56, opened 2020-12-18
return error from ParseIP("127.6.0.0%0")

josharian:
Found by fuzzing netaddr.ParseIP against wgcfg.ParseIP.


josharian:
This now contains a failing test. I think it'll be easier to fix the failing test after @bradfitz fixes #53.

danderson:
Looks like this got merged as part of something else, closing.

PR #57, opened 2020-12-18
Change representation to 20 bytes (32-bit) or 24 bytes (64-bit)

### LINKED ISSUE #53: Change representation of IP?

bradfitz:
The new representation is basically:

```go
   type IP struct {
      addr [16]byte
      zone *zone // always the same pointer for a given zone string
   }
```

See '"The zone pointer" alternative' comment from
https://github.com/inetaf/netaddr/issues/53#issuecomment-747849869

Fixes #53

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -21,6 +21,8 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+
+	"go4.org/intern"
 )
 
 // Simple constants copied to avoid importing math.
@@ -29,50 +31,43 @@ const (
 	maxUint16 = 1<<16 - 1
 )
 
+// o4 is the offset to the beginning of the IPv4 address within IP.a.
+const o4 = len(mapped4Prefix)
+
 // Sizes: (64-bit)
 //   net.IP:     24 byte slice header + {4, 16} = 28 to 40 bytes
 //   net.IPAddr: 40 byte slice header + {4, 16} = 44 to 56 bytes + zone length
-//   netaddr.IP: 16 byte interface + {4, 16, 24} = 20, 32, 40 bytes + zone length
+//   netaddr.IP: 24 bytes (zone is per-name singleton, shared across all users)
 
 // IP represents an IPv4 or IPv6 address (with or without a scoped
 // addressing zone), similar to Go's net.IP or net.IPAddr.
 //
 // Unlike net.IP or net.IPAddr, the netaddr.IP is a comparable value
 // type (it supports == and can be a map key) and is immutable.
-// Its memory representation ranges from 20 to 40 bytes, depending on
-// whether the underlying adddress is IPv4, IPv6, or IPv6 with a
-// zone. (This is smaller than the standard library's 28 to 56 bytes)
+// Its memory representation is 24 bytes in 64-bit machines (the same
+// size as a Go slice header) for both IPv4 and IPv6 address.
 type IP struct {
-	ipImpl
+	a [16]byte // IPv6 or IPv4-mapped IPv6 form of IPv4 addr
+
+	// z is a combination of the address family and the IPv6 zone.
+	//
+	// nil means invalid IP address (for the IP zero value).
+	// z4 means an IPv4 address.
+	// z6noz means an IPv6 address without a zone.
+	//
+	// Otherwise it's the interned zone name string.
+	z *intern.Value
 }
 
-// ipImpl is the interface representing either a v4addr, v6addr, v6ZoneAddr.
-type ipImpl interface {
-	is4() bool
-	is6() bool
-	is4in6() bool
-	as16() ip16
-	// prefix is the type-specific implementation of IP.Prefix.
-	prefix(uint8) (IPPrefix, error)
-	String() string
-}
+// z0, z4, and z6noz are sentinel IP.z values.
+// See the IP type's field docs.
+var (
+	z0    = (*intern.Value)(nil)
+	z4    = new(intern.Value)
+	z6noz = new(intern.Value)
+)
 
-type v4Addr [4]byte
-
-func (v4Addr) is4() bool    { return true }
-func (v4Addr) is6() bool    { return false }
-func (v4Addr) is4in6() bool { return false }
-func (ip v4Addr) as16() ip16 {
-	return ip16{
-		10: 0xff,
-		11: 0xff,
-		12: ip[0],
-		13: ip[1],
-		14: ip[2],
-		15: ip[3],
-	}
-}
-func (ip v4Addr) prefix(bits uint8) (IPPrefix, error) {
+func v4Prefix(ip [4]byte, bits uint8) (IPPrefix, error) {
 	if bits > 32 {
 		return IPPrefix{}, fmt.Errorf("netaddr: prefix length %d too large for IP address family", bits)
 	}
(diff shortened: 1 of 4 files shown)


bradfitz:
This should be done now. I'll add a few alloc tests still.

cc @josharian

netaddr.go
```diff
@@ -412,96 +332,118 @@ func (ip IP) Is6() bool {
...
+	zmu.Lock()
+	defer zmu.Unlock()
+
+	addr, ok := uniqZone[zoneName]
+	var z *zone
+	if ok {
+		z = (*zone)((unsafe.Pointer)(addr))
+	} else {
+		z = &zone{name: zoneName}
+		uniqZone[zoneName] = uintptr(unsafe.Pointer(z))
```
crawshaw, line 368: As this is outside the Go specification, I think:

1. It deserves commentary explaining that fact somewhere.
2. It would be good to add a check that could catch a potentially novel runtime and panic. A cheap one would be to put the zone uintptr in the *zone as an object header:

```
type zone struct {
    self uintptr
    ...
}
```

Then when using a zone, first check that `uintptr(unsafe.Pointer(z))==z.self` and panic if it doesn't.

josharian:
I'm very curious about doing this w/o unsafe. In any case, add a test like:

```
func TestWeakRefs(t *testing.T) {
	mustIP("fe80::1%eth0")
	runtime.GC()
	zmu.Lock()
	defer zmu.Unlock()
	if len(uniqZone) != 0 {
		t.Errorf("zone not cleaned up")
	}
}
```

netaddr.go
```diff
@@ -412,96 +323,82 @@ func (ip IP) Is6() bool {
...
+		ip.z = z4
 	}
-	a := ip.ipImpl.as16()
-	return IP{v4Addr{a[12], a[13], a[14], a[15]}}
+	return ip
 }
 
 // WithZone returns an IP that's the same as ip but with the provided
-// zone. If zone is empty, the zone is removed. If ip is an IPv4
+// zone. If zoneName is empty, the zone is removed. If ip is an IPv4
```
josharian, line 333: Now that there's no `zone` type, this can go back to being `zone` instead of `zoneName`.

netaddr.go
```diff
@@ -412,96 +323,82 @@ func (ip IP) Is6() bool {
...
+func (ip IP) WithZone(zoneName string) IP {
+	if !ip.Is6() {
 		return ip
 	}
-	switch ip := ip.ipImpl.(type) {
-	case v6Addr:
-		return IP{v6AddrZone{ip, zone}}
-	case v6AddrZone:
-		return IP{v6AddrZone{ip.v6Addr, zone}}
+	if zoneName == "" {
```
josharian, line 339: We could do `z6noz = intern.Get("")` as the initializer, at which point we could remove this check.

netaddr.go
```diff
@@ -412,96 +323,82 @@ func (ip IP) Is6() bool {
...
-		return ip.v6Addr[0] == 0xfe && ip.v6Addr[1] == 0x80
-	default:
-		panic("netaddr: unhandled ipImpl representation")
+	if ip.Is4() {
+		return ip.a[o4+0] == 169 && ip.a[o4+1] == 254
 	}
+	if ip.Is6() {
+		return ip.a[0] == 0xfe && ip.a[1] == 0x80
+	}
+	return false
```
josharian, line 356: Add a comment, please, like:

```go
return false // zero value
```

netaddr.go
```diff
@@ -412,96 +323,82 @@ func (ip IP) Is6() bool {
...
-		return string(ip.v6Addr[:len(v6Loopback)]) == v6Loopback
-	default:
-		panic("netaddr: unhandled ipImpl representation")
+	if ip.Is4() {
+		return ip.a[o4+0] == 127
 	}
+	if ip.Is6() {
+		return string(ip.a[:]) == v6Loopback
+	}
+	return false
```
josharian, line 368: Ditto here (add `// zero value`).

netaddr.go
```diff
@@ -96,13 +90,7 @@ const (
 	v6Loopback = "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01"
```
josharian, line 90: This could now be `v6Loopback = [16]byte{15: 1}`, I think.

(Can't do mapped4Prefix, until golang.org/issue/395 gets fixed, sadly.)

netaddr.go
```diff
@@ -412,96 +323,82 @@ func (ip IP) Is6() bool {
...
-		return string(ip[:len(v6Loopback)]) == v6Loopback
-	case v6AddrZone:
-		return string(ip.v6Addr[:len(v6Loopback)]) == v6Loopback
-	default:
-		panic("netaddr: unhandled ipImpl representation")
+	if ip.Is4() {
+		return ip.a[o4+0] == 127
 	}
+	if ip.Is6() {
+		return string(ip.a[:]) == v6Loopback
```
josharian, line 366: If you change this to `[16]byte`, you can drop the string conversion and use == directly.

netaddr.go
```diff
@@ -29,50 +31,43 @@ const (
 	maxUint16 = 1<<16 - 1
 )
 
+// o4 is the offset to the beginning of the IPv4 address within IP.a.
+const o4 = 12
```
josharian, line 35: Consider `const o4 = len(mapped4Prefix)`. Then you can also `s/len(mapped4Prefix)/o4/` throughout.

netaddr.go
```diff
@@ -224,17 +169,12 @@ func ParseIP(s string) (IP, error) {
...
 		}
 	}
-	var v6 v6Addr
-	copy(v6[:], ipa.IP.To16())
-	if ipa.Zone != "" {
-		return IP{v6AddrZone{v6, ipa.Zone}}, nil
-	}
-	return IP{v6}, nil
+	var a16 [16]byte
+	copy(a16[:], ipa.IP.To16())
```
josharian, line 176: It makes me sad what terrible code the compiler generates for this. Not important enough to be worth changing here (the fix is to do two binary.NativeEndian.[Put]Uint64 calls), just...sad.

netaddr.go
```diff
@@ -293,51 +229,38 @@ func (ip IP) IsZero() bool { return ip == IP{} }
...
 
 // Zone returns ip's IPv6 scoped addressing zone, if any.
 func (ip IP) Zone() string {
-	if v6z, ok := ip.ipImpl.(v6AddrZone); ok {
-		return v6z.zone
+	if ip.z == nil {
+		return ""
 	}
-	return ""
+	zoneName, _ := ip.z.Get().(string)
```
josharian, line 246: We shouldn't make `intern.Value.Get` return nil when it has a nil receiver, right?

netaddr.go
```diff
@@ -545,17 +434,24 @@ func (ip IP) As4() [4]byte {
 // Note that unlike the Go standard library's IP.String method,
 // IP4-mapped IPv6 addresses do not format as dotted decimals.
 func (ip IP) String() string {
```
josharian, line 436: I assume we'll come back and optimize this method more later.

josharian (approved):

netaddr.go
```diff
@@ -115,72 +103,29 @@ func (ip v6Addr) prefix(bits uint8) (IPPrefix, error) {
...
 
 // IPv6Unspecified returns the IPv6 unspecified address ::.
-func IPv6Unspecified() IP { return ipv6Unspecified }
+func IPv6Unspecified() IP { return IPv6Raw([16]byte{}) }
 
 // IPv4 returns the IP of the IPv4 address a.b.c.d.
 func IPv4(a, b, c, d uint8) IP {
-	return IP{v4Addr{a, b, c, d}}
+	return IP{
+		a: [16]byte{10: 0xff, 11: 0xff, 12: a, 13: b, 14: c, 15: d},
```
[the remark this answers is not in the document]
bradfitz: I wavered on that too. It looks ugly all on one line (gofmt adds spaces around the `+` operators), but it looks fine on a line per field. Done.

netaddr.go
bradfitz: Done.

netaddr.go
bradfitz: done

netaddr.go
bradfitz: Done.

netaddr.go
bradfitz: Done.

netaddr.go
bradfitz: I think I'd rather catch mistakes and explode if you use a nil.

netaddr.go
bradfitz: yeah, those are the commented out alloc TODOs in the test

netaddr.go
bradfitz: File a Go bug?

netaddr.go
bradfitz: Done.

mdlayher:
Really nice work. I will update my code that uses this package.

PR #58, opened 2020-12-18
extract and document handling of zones

josharian:
Pull all the zone stuff into its own file.
Document in detail how it works,
and why it looks the way it does.

Add future build breakage to force us to recertify
that the package still works.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -17,10 +17,12 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
+	"unsafe"
 )
 
 // Simple constants copied to avoid importing math.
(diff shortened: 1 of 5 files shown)


zone.go
```diff
@@ -0,0 +1,124 @@
+// Copyright 2020 The Inet.Af AUTHORS. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !go1.19
```
danderson, line 5: Why 1.19, rather than "the next release" (so, 1.17-ish) ?

zone.go
```diff
@@ -0,0 +1,124 @@
...
+	gen int64
+}
+
+var (
+	z0    = (*zone)(nil)
+	z4    = new(zone)
+	z6noz = new(zone)
+)
+
+// zmu guards uniqZone, a weakref map of *zones by zoneName.
```
danderson, line 33: Mention here that zmu also guards fields within zone. It's mentioned above, but is unusual enough that I think it deserves an xref here.

zone.go
```diff
@@ -0,0 +1,124 @@
...
+// internZone returns an interned zone for zoneName.
+//
+// We intern zones to guarantee that two zone pointers are equal
+// iff their names are equal, so that IPs can be compared, used as map keys, etc.
+//
+// Interning is simple if you don't require that unused zones be garbage collectable.
+// But we do require that; we don't want to be DOS vector.
+// We do this by using a uintptr to hide our zone pointer from the garbage collector,
+// and using a finalizer to eliminate our zone pointer when no other code is using it.
+// (The checkptr runtime check prevents exactly these kinds of shenanigans; we disable it.)
```
danderson, line 48: Can you say more words as to what exactly checkptr enforces/detects, that we're disabling? At first glance I don't understand what exactly about these shenanigans make it angry.

zone.go
```diff
@@ -0,0 +1,124 @@
...
+// Another, later, finalizer will take care of that.
+// The zone generation field is protected by zmu, providing a consistent view.
+//
+// @josharian has a mild lingering concern about this approach.
+// It is possible to for the runtime to concurrently decide it needs to _execute_ a finalizer and
+// also _remove_ the need for that finalizer to run, because a new reference has appeared.
+// It is possible that this could cause a data race in the runtime.
+// This is not a normal thing to have happen; it requires unsafe hiding of a pointer in a uintptr.
+// It thus might not be tested for or protected against in the runtime.
+// Hopefully this will not prove to be a problem in practice.
```
danderson, line 77: How would we detect this happening? Is there any way for us to panic and explode at runtime, rather than end up with another iteration of "windows memory corruption"?

zone_future.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 The Inet.Af AUTHORS. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.19
```
danderson, line 5: Same question re 1.19 vs. 1.17. Do we feel okay that the runtime won't mutate too much until summer 2022?

zone.go
josharian: @bradfitz suggested 1.19. I'm not sure why.

zone.go
josharian: Done.

zone.go
josharian: Quoting the 1.14 release notes, checkptr is "a compile-time option for adding instrumentation to check that Go code is following unsafe.Pointer safety rules dynamically."

Package unsafe says:

>  The remaining patterns enumerate the only valid conversions from uintptr to Pointer. 
>  (3) Conversion of a Pointer to a uintptr and back, with arithmetic. 
>  (4) Conversion of a Pointer to a uintptr when calling syscall.Syscall. 
>  (5) Conversion of the result of reflect.Value.Pointer or reflect.Value.UnsafeAddr from uintptr to Pointer. 
>  (6) Conversion of a reflect.SliceHeader or reflect.StringHeader Data field to or from Pointer. 

Storing a uintptr and later making it back into an unsafe.Pointer is the golden example of what you're not supposed to do with package unsafe.

Which (if any) of these words would be helpful to add to the comments here? They're mostly about checkptr and the unsafe rules, not about our code.

zone.go
josharian: > How would we detect this happening?

I don't know of a way.

> Is there any way for us to panic and explode at runtime, rather than end up with another iteration of "windows memory corruption"?

Not that I know of.

It is worth stressing that this is a theoretical possibility. The runtime might be robust to this in practice, particularly since the function pointer is a single word. I started going through the implementation to try to find out by reading the code, but the code is, um, intricate.

bradfitz:
Did you see that I'd previously pulled out all the shenanigans into https://github.com/bradfitz/litecmp ? (temporary location... `go4.org/foo` instead?)

It might be better having all the unsafe in a different package to keep it easier to audit.

josharian:
Oh, I thought that was intended to be for demo/discussion. Almost everything meaningful I changed in the first commit here was words—feel free to just steal them and edit and apply to whatever home you see fit. If you want package path ideas,  maybe go4.org/intern. Then could have intern.Strings/intern.Bytes (stolen from github.com/josharian/intern) and intern.Comparable (better name?).

netaddr.go
```diff
@@ -431,7 +431,7 @@ func (ip IP) String() string {
 	}
 	if ip.Is4in6() {
 		a4 := ip.As4()
-		return fmt.Sprintf("::ffff:%x%02d:%x%02x", a4[0], a4[1], a4[2], a4[3])
+		return fmt.Sprintf("::ffff:%x:%x", uint16(a4[0])<<8|uint16(a4[1]), uint16(a4[2])<<8|uint16(a4[3]))
```
bradfitz, line 434: whoops, thanks. This was a late addition to the change and I thought it was already covered by tests so I got over-confident I guess.

PR #61, opened 2020-12-23
Add extractPrefix to IPPrefix and IPSet

majst01:
Hi,

Thanks for this library, i enjoy using it as a more convenient api compared to the stdlib variant.
I have a use case where i need to extract a prefix with given bit length from a existing larger prefix/set.

If you find this addition useful, please consider it.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -716,7 +716,7 @@ func splitIPPort(s string) (ip, port string, v6 bool, err error) {
 		if len(ip) < 2 || ip[len(ip)-1] != ']' {
 			return "", "", false, errors.New("missing ]")
 		}
-		ip = ip[1:len(ip)-1]
+		ip = ip[1 : len(ip)-1]
 		v6 = true
 	}
 
@@ -1411,6 +1411,38 @@ func (s *IPSet) AddRange(r IPRange) {
 
 // Remove removes ip from the set s.
 func (s *IPSet) Remove(ip IP) { s.RemoveRange(IPRange{ip, ip}) }
+
+// RemoveFreePrefix removes and returns a Prefix of length bits from the IPSet.
+func (s *IPSet) RemoveFreePrefix(bitLen uint8) (p IPPrefix, ok bool) {
+	prefixes := s.Prefixes()
+	if len(prefixes) == 0 {
+		return IPPrefix{}, false
+	}
+	existingPrefixes := make(map[uint8]IPPrefix)
+	for _, prefix := range prefixes {
+		existingPrefixes[prefix.Bits] = prefix
+	}
+	exactMatch, ok := existingPrefixes[bitLen]
+	if ok {
+		s.RemovePrefix(exactMatch)
+		return exactMatch, true
+	}
+
+	nextBiggerPrefix, ok := existingPrefixes[bitLen-1]
+	if !ok {
+		if len(prefixes) < 1 {
+			return IPPrefix{}, false
+		}
+		pfx := prefixes[0]
+		prefix := IPPrefix{IP: pfx.IP, Bits: bitLen}
+		s.RemovePrefix(prefix)
+		return prefix, true
+	}
+
+	prefix := IPPrefix{IP: nextBiggerPrefix.IP, Bits: bitLen}
+	s.RemovePrefix(prefix)
+	return prefix, true
+}
 
 // RemovePrefix removes p's range from s.
 func (s *IPSet) RemovePrefix(p IPPrefix) { s.RemoveRange(p.Range()) }
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -1677,7 +1677,54 @@ func pxv(cidrStrs ...string) (out []IPPrefix) {
 	}
 	return
 }
-
+func TestIPSetRemoveFreePrefix(t *testing.T) {
+	pfx := mustIPPrefix
+	tests := []struct {
+		name         string
+		f            func(s *IPSet)
+		b            uint8
+		wantPrefix   IPPrefix
+		wantPrefixes []IPPrefix
+		wantOK       bool
+	}{
+		{
+			name: "cut in half",
+			f: func(s *IPSet) {
+				s.AddPrefix(pfx("10.0.0.0/8"))
+			},
+			b:            9,
+			wantPrefix:   pfx("10.0.0.0/9"),
+			wantPrefixes: pxv("10.128.0.0/9"),
+			wantOK:       true,
+		},
+		{
+			name: "on prefix left",
+			f: func(s *IPSet) {
+				s.AddPrefix(pfx("10.0.0.0/8"))
+				s.RemovePrefix(pfx("10.0.0.0/9"))
+			},
+			b:            9,
+			wantPrefix:   pfx("10.128.0.0/9"),
+			wantPrefixes: nil,
+			wantOK:       true,
+		},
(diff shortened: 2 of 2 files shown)


bradfitz:
I'm not exactly clear on what this is supposed to do based on its docs. Is this the normal name of this operation? Can you point to docs online about it?

As for implementation, though: the repeated calls to `.Range()` and `.Prefixes()` aren't cheap. You should pull those out into variables and minimize how often you need to call those methods.

/cc @mdlayher too.

majst01:
Hi @bradfitz,

tried to clarify the usage and unexported the lesser useful implementation.

I am the author of https://github.com/metal-stack/go-ipam which is the IPAM backend implementation we use for https://github.com/metal-stack which is a bare metal server provisioning.

In the current implementation of go-ipam we dont support ipv6 and variable child prefix length. I added ipv6 support by replacing our bad ipv4 math with the netaddr math. And for the child prefix calculation i use your IPSet.Prefixes function to divide a given Prefix in smaller chunks: https://github.com/metal-stack/go-ipam/pull/36

We use these child prefixes to give every tenant its own address space for his servers, and this address space is part of a existing bigger network.

Hope this helps and thank for looking at my PR.

mdlayher:
I don't have strong feelings about this at the moment as I haven't made any use of the IPSet type in my own code.

I've previously used https://github.com/mikioh/ipaddr for some of my IPAM-like applications, but my netaddr code these days really just cares about /64s.

danderson:
I'm on the fence about this. It means that netaddr starts to get into the IPAM business, which is not what IPSet was originally made for. Then again, prefix extraction is a bunch of subtle code that is annoying to write, so having a single well-debugged implementation seems nice. If I were still developing MetalLB, I would probably use this netaddr method.

Reviewing on the tentative assumption that we'll accept the new API.

netaddr.go
```diff
@@ -1086,6 +1086,78 @@ func (s *IPSet) RemoveSet(b *IPSet) {
...
+// childPrefix, _ := set.ExtractPrefix(18)
+//
+// childPrefix will be: "192.168.0.0/18"
+func (s *IPSet) ExtractPrefix(length uint8) (IPPrefix, error) {
+	prefixes := s.Prefixes()
+	if len(prefixes) == 0 {
+		return IPPrefix{}, fmt.Errorf("no more child prefixes contained in prefix pool")
+	}
+	existingPrefixes := make(map[uint8]IPPrefix)
+	for _, prefix := range prefixes {
```
danderson, line 1130: I think you can make this more efficient:

```
var bestFit IPPrefix
for _, prefix := range prefixes {
  if prefix.Bits > length {
    continue
  }
  if bestFit.IP.IsZero() || prefix.Bits > bestFit.Bits {
    bestFit = prefix
  }
  if bestFit.Bits == length {
    // exact match, done
    break
  }
}
```

At the end of that iteration, bestFit will be the best prefix to use - either an exact match, or the closest in size - or bestFit.IP.IsZero()==true if no suitable prefix exists. As a bonus, it exits early if it finds an exact match, and you can remove the explicit test for len()==0 above.

danderson (commented):
Looks good in general, but I have suggestions for how to rewrite it to be more efficient.

An open question I have: IPAMs generally offer multiple IP allocation policies, and people who use IPAMs tend to care deeply about these things. This implementation implements one specific allocation strategy (best fit, with no preference for how top-level prefixes get consumed), which I would summarize as "don't care, I just want an IPPrefix".

Is it right for us to give that strategy priority over all others? Strictly speaking it's not a huge deal, because IPSet exposes enough primitives to let people implement other strategies externally, but then why are we giving one specific strategy prominent placement in netaddr?

"Because it's a good implementation for people who don't care about IPAM allocation strategies" is probably an okay answer, but I'm curious what others think.

netaddr.go
```diff
@@ -1086,6 +1086,78 @@ func (s *IPSet) RemoveSet(b *IPSet) {
 	}
 }
 
+// extractPrefix will return a Prefix with a bitlength of length which is part of the given Prefix.
+func (p IPPrefix) extractPrefix(length uint8) (IPPrefix, error) {
```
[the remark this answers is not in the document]
majst01: Ha, right, done

netaddr.go
```diff
@@ -1086,6 +1086,78 @@ func (s *IPSet) RemoveSet(b *IPSet) {
...
+	}
+	for _, srp := range srpfx {
+		if srp.Bits == length {
+			return srp, nil
+		}
+	}
+	return IPPrefix{}, fmt.Errorf("no prefix with length:%d found in %s", length, p)
+}
+
+// ExtractPrefix will return a Prefix with a bitlength of length which is part of the given IPSet.
```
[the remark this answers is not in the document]
majst01: done

netaddr.go
```diff
@@ -1086,6 +1086,78 @@ func (s *IPSet) RemoveSet(b *IPSet) {
...
+// The Prefix is also removed from the IPSet.
+//
+// Unlike RemovePrefix this function will not require you to give a Prefix, instead you
```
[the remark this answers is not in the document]
majst01: done as well

netaddr.go
```diff
@@ -1086,6 +1086,78 @@ func (s *IPSet) RemoveSet(b *IPSet) {
...
+// Example usage:
+// pfx,_ := netaddr.ParsePrefix("192.168.0.0/16")
+// set.AddPrefix(pfx)
+// childPrefix, _ := set.ExtractPrefix(18)
+//
+// childPrefix will be: "192.168.0.0/18"
+func (s *IPSet) ExtractPrefix(length uint8) (IPPrefix, error) {
+	prefixes := s.Prefixes()
+	if len(prefixes) == 0 {
+		return IPPrefix{}, fmt.Errorf("no more child prefixes contained in prefix pool")
```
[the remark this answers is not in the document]
majst01: also

netaddr.go
majst01: Unsure, don't understand really, bestFit is not set in you example code.

bradfitz:
Naming-wise, maybe `RemoveFreePrefix(bitLen int) (p IPPrefix, ok bool)` ? Or `RemovedSomePrefixOfLength`, etc.

Then it's obvious it removes it, and it matches the other `RemoveXXX` methods.

netaddr.go
danderson: It's set in the loop, in the second if. It starts out as the zero value, then gets overridden by the first "big enough" prefix, and gets updated as more precise prefixes are found.

majst01:
Renamed to what @bradfitz suggested, inlined simple func, force pushed to one commit.

mdlayher:
Looks like this needs a rebase.

netaddr.go
```diff
@@ -1132,6 +1132,39 @@ func (s *IPSet) AddRange(r IPRange) {
 	s.in = append(s.in, r)
 }
 
+// RemoveFreePrefix removes and returns a Prefix of length bits from the IPSet.
+// The Prefix is also removed from the IPSet.
```
bradfitz, line 1136: this sentence seems redundant with the two "remove" words on the previous line (which, while redundant, are conventional)

majst01: done and rebased

danderson (approved):

majst01:
Thanks guys !

PR #62, opened 2020-12-23
Test that Prior.Next and Next.Prior return same IP

jawnsy:
Thanks for creating this library! This is something that I had wished existed before, so I'm glad you've taken the initiative to create this. I want to contribute to this effort and thought that adding some test cases would be a good way to go about it. If you have hints for other "good first issues" to tackle, I'd appreciate the direction!

As long as the Prior or Next address is valid, then doing the
inverse operation should get us the IP we started with.

This also adds a few test cases for localhost addresses.

### DIFF
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -1790,11 +1790,13 @@ func TestIPNextPrior(t *testing.T) {
 	}{
 		{mustIP("10.0.0.1"), mustIP("10.0.0.2"), mustIP("10.0.0.0")},
 		{mustIP("10.0.0.255"), mustIP("10.0.1.0"), mustIP("10.0.0.254")},
+		{mustIP("127.0.0.1"), mustIP("127.0.0.2"), mustIP("127.0.0.0")},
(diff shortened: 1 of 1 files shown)


josharian:
Nice test. I threw it into the fuzz function I have brewing, and it immediately spat out this problematic test case:

```
ip= ::%0 .next= ::1 .next.prior= ::
panic: .Next.Prior did not round trip
```

Looks like .Next and .Prior don't preserve IPv6 zones.

mdlayher (approved):
LGTM

josharian:
@mdlayher do we need a Signed-Off-By here? Not sure how picky Brad is about licensing...

mdlayher:
I dunno for sure! I've had it turned on by default for a while in my commits. Probably good to keep it up.

jawnsy:
Rebased and added signoff 😄 

I should look into turning it on by default, that's an excellent idea...

josharian:
> Edit: It might also be a good idea to add a GitHub Action to enforce this if desired?

Definitely. I'll file an issue.

PR #63, opened 2020-12-24
Change the representation of IP to a pair of uint64s.

danderson:
This allows IP manipulation operations to operate on IPs as two registers, which empirically
leads to significant speedups, in particular on OOO superscalars where the two halves can
be processed in parallel.

```
name                              old time/op    new time/op    delta
StdIPv4-8                            146ns ± 2%     141ns ± 2%    -3.42%  (p=0.016 n=5+5)
IPv4-8                               120ns ± 1%     107ns ± 2%   -10.65%  (p=0.008 n=5+5)
IPv4_inline-8                        120ns ± 0%     118ns ± 1%    -1.67%  (p=0.016 n=4+5)
StdIPv6-8                            211ns ± 2%     215ns ± 1%    +2.18%  (p=0.008 n=5+5)
IPv6-8                               281ns ± 1%     252ns ± 1%   -10.19%  (p=0.008 n=5+5)
IPv4Contains-8                      11.8ns ± 4%     4.7ns ± 2%   -60.00%  (p=0.008 n=5+5)
ParseIPv4-8                         68.1ns ± 4%    78.8ns ± 1%   +15.74%  (p=0.008 n=5+5)
ParseIPv6-8                          419ns ± 1%     409ns ± 0%    -2.40%  (p=0.016 n=4+5)
StdParseIPv4-8                      73.7ns ± 1%    88.8ns ± 2%   +20.50%  (p=0.008 n=5+5)
StdParseIPv6-8                       132ns ± 2%     134ns ± 1%      ~     (p=0.079 n=5+5)
IPPrefixMasking/IPv4_/32-8          36.3ns ± 3%     4.8ns ± 4%   -86.72%  (p=0.008 n=5+5)
IPPrefixMasking/IPv4_/17-8          39.0ns ± 0%     4.8ns ± 3%   -87.78%  (p=0.008 n=5+5)
IPPrefixMasking/IPv4_/0-8           36.9ns ± 2%     4.8ns ± 4%   -87.07%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_/128-8         32.7ns ± 1%     4.7ns ± 2%   -85.47%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_/65-8          39.8ns ± 1%     4.7ns ± 1%   -88.13%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_/0-8           40.7ns ± 1%     4.7ns ± 2%   -88.41%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/128-8     136ns ± 3%       5ns ± 2%   -96.53%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/65-8      142ns ± 2%       5ns ± 1%   -96.65%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/0-8       143ns ± 2%       5ns ± 3%   -96.67%  (p=0.008 n=5+5)
IPSetFuzz-8                         22.7µs ± 2%    16.4µs ± 2%   -27.84%  (p=0.008 n=5+5)

name                              old alloc/op   new alloc/op   delta
StdIPv4-8                            16.0B ± 0%     16.0B ± 0%      ~     (all equal)
IPv4-8                               0.00B          0.00B           ~     (all equal)
IPv4_inline-8                        0.00B          0.00B           ~     (all equal)
StdIPv6-8                            16.0B ± 0%     16.0B ± 0%      ~     (all equal)
IPv6-8                               16.0B ± 0%     16.0B ± 0%      ~     (all equal)
IPv4Contains-8                       0.00B          0.00B           ~     (all equal)
ParseIPv4-8                          0.00B          0.00B           ~     (all equal)
ParseIPv6-8                          48.0B ± 0%     48.0B ± 0%      ~     (all equal)
StdParseIPv4-8                       16.0B ± 0%     16.0B ± 0%      ~     (all equal)
StdParseIPv6-8                       16.0B ± 0%     16.0B ± 0%      ~     (all equal)
IPPrefixMasking/IPv4_/32-8           0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv4_/17-8           0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv4_/0-8            0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv6_/128-8          0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv6_/65-8           0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv6_/0-8            0.00B          0.00B           ~     (all equal)
IPPrefixMasking/IPv6_zone_/128-8     16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/65-8      16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/0-8       16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
IPSetFuzz-8                         2.60kB ± 0%    2.60kB ± 0%      ~     (p=1.000 n=5+5)

name                              old allocs/op  new allocs/op  delta
StdIPv4-8                             1.00 ± 0%      1.00 ± 0%      ~     (all equal)
IPv4-8                                0.00           0.00           ~     (all equal)
IPv4_inline-8                         0.00           0.00           ~     (all equal)
StdIPv6-8                             1.00 ± 0%      1.00 ± 0%      ~     (all equal)
IPv6-8                                1.00 ± 0%      1.00 ± 0%      ~     (all equal)
IPv4Contains-8                        0.00           0.00           ~     (all equal)
ParseIPv4-8                           0.00           0.00           ~     (all equal)
ParseIPv6-8                           3.00 ± 0%      3.00 ± 0%      ~     (all equal)
StdParseIPv4-8                        1.00 ± 0%      1.00 ± 0%      ~     (all equal)
StdParseIPv6-8                        1.00 ± 0%      1.00 ± 0%      ~     (all equal)
IPPrefixMasking/IPv4_/32-8            0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv4_/17-8            0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv4_/0-8             0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv6_/128-8           0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv6_/65-8            0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv6_/0-8             0.00           0.00           ~     (all equal)
IPPrefixMasking/IPv6_zone_/128-8      1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/65-8       1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
IPPrefixMasking/IPv6_zone_/0-8        1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
IPSetFuzz-8                           33.0 ± 0%      33.0 ± 0%      ~     (all equal)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -13,7 +13,7 @@
 package netaddr // import "inet.af/netaddr"
 
 import (
-	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"net"
@@ -25,14 +25,8 @@ import (
 	"go4.org/intern"
 )
 
-// Simple constants copied to avoid importing math.
-const (
-	maxUint8  = 1<<8 - 1
-	maxUint16 = 1<<16 - 1
-)
-
-// o4 is the offset to the beginning of the IPv4 address within IP.a.
-const o4 = len(mapped4Prefix)
+// copied to avoid importing math.
+const maxUint16 = 1<<16 - 1
 
 // Sizes: (64-bit)
 //   net.IP:     24 byte slice header + {4, 16} = 28 to 40 bytes
@@ -47,7 +41,22 @@ const o4 = len(mapped4Prefix)
 // Its memory representation is 24 bytes on 64-bit machines (the same
 // size as a Go slice header) for both IPv4 and IPv6 address.
 type IP struct {
-	a [16]byte // IPv6 or IPv4-mapped IPv6 form of IPv4 addr
+	// hi and lo are the bits of an IPv6 address. If z==z4, hi and lo
+	// contain the IPv4-mapped IPv6 address.
+	//
+	// hi and lo are constructed by interpreting a 16-byte IPv6
+	// address as a big-endian 128-bit number. The most significant
+	// bits of that number go into hi, the rest into lo.
+	//
+	// For example, 0011:2233:4455:6677:8899:aabb:ccdd:eeff is stored as:
+	//  hi = 0x0011223344556677
+	//  lo = 0x8899aabbccddeeff
+	//
+	// We store IPs like this, rather than as [16]byte, because it
+	// turns most operations on IPs into arithmetic and bit-twiddling
+	// operations on 64-bit registers, which is much faster than
+	// bytewise processing.
+	hi, lo uint64
 
 	// z is a combination of the address family and the IPv6 zone.
 	//
@@ -67,64 +76,18 @@ var (
 	z6noz = new(intern.Value)
 )
 
-func v4Prefix(ip [4]byte, bits uint8) (IPPrefix, error) {
-	if bits > 32 {
-		return IPPrefix{}, fmt.Errorf("netaddr: prefix length %d too large for IP address family", bits)
-	}
-	skip, partial := int(bits/8), bits%8
-	if partial != 0 {
-		ip[skip] = ip[skip] & ^byte(0xff>>partial)
-		skip++
-	}
-	for i := skip; i < 4; i++ {
-		ip[i] = 0
-	}
-	return IPPrefix{IPv4(ip[0], ip[1], ip[2], ip[3]), bits}, nil
-}
-
-const (
-	// mapped4Prefix are the 12 leading bytes in a IPv4-mapped IPv6 address.
-	mapped4Prefix = "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff"
-)
-
-// v6Loopback is the IPv6 loopback address.
-var v6Loopback = [16]byte{15: 1}
-
-func v6Prefix(ip [16]byte, bits uint8) (IPPrefix, error) {
-	if bits > 128 {
-		return IPPrefix{}, fmt.Errorf("netaddr: prefix length %d too large for IP address family", bits)
-	}
-	skip, partial := int(bits/8), bits%8
-	if partial != 0 {
-		ip[skip] = ip[skip] & ^byte(0xff>>partial)
-		skip++
-	}
-	b := ip[skip:]
-	for i := range b {
-		b[i] = 0
-	}
-	return IPPrefix{IPv6Raw(ip), bits}, nil
-}
-
 // IPv6LinkLocalAllNodes returns the IPv6 link-local all nodes multicast
 // address ff02::1.
 func IPv6LinkLocalAllNodes() IP { return IPv6Raw([16]byte{0: 0xff, 1: 0x02, 15: 0x01}) }
 
 // IPv6Unspecified returns the IPv6 unspecified address ::.
-func IPv6Unspecified() IP { return IPv6Raw([16]byte{}) }
+func IPv6Unspecified() IP { return IP{z: z6noz} }
 
 // IPv4 returns the IP of the IPv4 address a.b.c.d.
 func IPv4(a, b, c, d uint8) IP {
 	return IP{
-		z: z4,
-		a: [16]byte{
-			10:     0xff,
-			11:     0xff,
-			o4 + 0: a,
-			o4 + 1: b,
-			o4 + 2: c,
-			o4 + 3: d,
-		},
+		lo: 0xffff00000000 | uint64(a)<<24 | uint64(b)<<16 | uint64(c)<<8 | uint64(d),
+		z:  z4,
 	}
 }
 
@@ -132,19 +95,29 @@ func IPv4(a, b, c, d uint8) IP {
 // without an implicit Unmap call to unmap any v6-mapped IPv4
 // address.
 func IPv6Raw(addr [16]byte) IP {
-	return IP{a: addr, z: z6noz}
+	return IP{
+		hi: binary.BigEndian.Uint64(addr[:8]),
+		lo: binary.BigEndian.Uint64(addr[8:]),
+		z:  z6noz,
+	}
+}
+
+// ipv6Slice is like IPv6Raw, but operates on a 16-byte slice. Assumes
+// slice is 16 bytes, caller must enforce this.
+func ipv6Slice(addr []byte) IP {
+	return IP{
+		hi: binary.BigEndian.Uint64(addr[:8]),
+		lo: binary.BigEndian.Uint64(addr[8:]),
+		z:  z6noz,
+	}
 }
 
 // IPFrom16 returns the IP address given by the bytes in addr,
 // unmapping any v6-mapped IPv4 address.
 //
-// It is equivalent to calling IPv6Raw(addr).Unmap() but slightly more
-// efficient.
+// It is equivalent to calling IPv6Raw(addr).Unmap().
 func IPFrom16(addr [16]byte) IP {
-	if string(addr[:o4]) == mapped4Prefix {
-		return IPv4(addr[o4+0], addr[o4+1], addr[o4+2], addr[o4+3])
-	}
-	return IPv6Raw(addr)
+	return IPv6Raw(addr).Unmap()
 }
 
 // ParseIP parses s as an IP address, returning the result. The string
@@ -187,9 +160,7 @@ func ParseIP(s string) (IP, error) {
 			return IPv4(ip4[0], ip4[1], ip4[2], ip4[3]), nil
 		}
 	}
-	var a16 [16]byte
-	copy(a16[:], ipa.IP.To16())
-	return IPv6Raw(a16).WithZone(ipa.Zone), nil
+	return ipv6Slice(ipa.IP.To16()).WithZone(ipa.Zone), nil
 }
 
 // MustParseIP calls ParseIP(s) and panics on error.
@@ -239,18 +210,11 @@ func parseIPv4(s string) (ip IP, ok bool) {
 // To convert a standard library IP without the implicit unmapping, use
 // FromStdIPRaw.
 func FromStdIP(std net.IP) (ip IP, ok bool) {
-	if len(std) == 16 && string(std[:o4]) == mapped4Prefix {
-		std = std[o4:]
+	ret, ok := FromStdIPRaw(std)
+	if ret.Is4in6() {
+		ret.z = z4
 	}
-	switch len(std) {
-	case 4:
-		return IPv4(std[0], std[1], std[2], std[3]), true
-	case 16:
-		var a [16]byte
-		copy(a[:], std)
-		return IPFrom16(a), true
-	}
-	return IP{}, false
+	return ret, ok
 }
 
 // FromStdIPRaw returns an IP from the standard library's IP type.
@@ -262,11 +226,25 @@ func FromStdIPRaw(std net.IP) (ip IP, ok bool) {
 	case 4:
 		return IPv4(std[0], std[1], std[2], std[3]), true
 	case 16:
-		var a [16]byte
-		copy(a[:], std)
-		return IPv6Raw(a), true
+		return ipv6Slice(std), true
 	}
 	return IP{}, false
+}
+
+// v4 returns the i'th byte of ip. If ip is not an IPv4, v4 returns
+// unspecified garbage.
+func (ip IP) v4(i uint8) uint8 {
+	return uint8(ip.lo>>((3-i)*8))
+}
+
+// v6 returns the i'th byte of ip. If ip is an IPv4 address, this
+// accesses the IPv4-mapped IPv6 address form of the IP.
+func (ip IP) v6(i uint8) uint8 {
+	if i >= 8 {
+		return uint8(ip.lo>>((15-i)*8))
+	} else {
+		return uint8(ip.hi>>((7-i)*8))
+	}
 }
 
 // IsZero reports whether ip is the zero value of the IP type.
@@ -309,16 +287,27 @@ func (ip IP) Compare(ip2 IP) int {
 	if f1 > f2 {
 		return 1
 	}
-	c := bytes.Compare(ip.a[:], ip2.a[:])
-	if c == 0 && ip.Is6() {
+	if ip.hi < ip2.hi {
+		return -1
+	}
+	if ip.hi > ip2.hi {
+		return 1
+	}
+	if ip.lo < ip2.lo {
+		return -1
+	}
+	if ip.lo > ip2.lo {
+		return 1
+	}
+	if ip.Is6() {
 		za, zb := ip.Zone(), ip2.Zone()
 		if za < zb {
-			c = -1
+			return -1
 		} else if za > zb {
-			c = 1
+			return 1
 		}
 	}
-	return c
+	return 0
 }
 
 // Less reports whether ip sorts before ip2.
@@ -338,7 +327,8 @@ func (ip IP) ipZone(reuse net.IP) (stdIP net.IP, zone string) {
 		a4 := ip.As4()
 		return append(base, a4[:]...), ""
 	default:
-		return append(base, ip.a[:]...), ip.Zone()
+		a16 := ip.As16()
+		return append(base, a16[:]...), ip.Zone()
 	}
 }
 
@@ -359,7 +349,7 @@ func (ip IP) Is4() bool {
 
 // Is4in6 reports whether ip is an IPv4-mapped IPv6 address.
 func (ip IP) Is4in6() bool {
-	return ip.z != z0 && ip.z != z4 && string(ip.a[:o4]) == mapped4Prefix
+	return ip.Is6() && ip.hi == 0 && ip.lo>>32 == 0xffff
 }
 
 // Is6 reports whether ip is an IPv6 address, including IPv4-mapped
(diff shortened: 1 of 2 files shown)


netaddr.go
```diff
@@ -13,7 +13,7 @@
 package netaddr // import "inet.af/netaddr"
 
 import (
-	"bytes"
+	"encoding/binary"
```
danderson, line 16: We could avoid this import pretty easily if we felt the need. We're invoking pretty trivial functions that get inlined, nothing complex.

josharian: This import doesn't worry me.

netaddr.go
```diff
@@ -25,15 +25,13 @@ import (
 	"go4.org/intern"
 )
 
-// Simple constants copied to avoid importing math.
 const (
-	maxUint8  = 1<<8 - 1
+	// copied to avoid importing math.
 	maxUint16 = 1<<16 - 1
+	// mask4In6 is the mask of ip.lo indicating an IPv4-mapped IPv6 address.
+	mask4In6 = 0xffff00000000
```
josharian, line 32: If this is the problematic bitmap that varies by endianness, the easiest fix is to put the different constants in build-tag-protected files. Or have an `IsLittleEndian` const that is 0 or 1, and do multiply-and-add tricks to construct the correct value from it (`IsLittleEndian*littleEndianBitmap+(1-IsLittleEndian)*bigEndianBitmap`). If you planned to use github.com/josharian/native, I'd be happy to add such a constant there. We could also move that to go4.org if we want everything in one place.

Or we could accept the BSWAPQ, of course.

netaddr.go
```diff
@@ -47,7 +45,7 @@ const o4 = len(mapped4Prefix)
 // Its memory representation is 24 bytes on 64-bit machines (the same
 // size as a Go slice header) for both IPv4 and IPv6 address.
 type IP struct {
-	a [16]byte // IPv6 or IPv4-mapped IPv6 form of IPv4 addr
+	hi, lo uint64 // IPv6 or IPv4-mapped IPv6 form of IPv4 addr
```
josharian, line 48: Say something about endianness here. Maybe write out an example IPv6 address and indicate what the hi and lo values are for it (in hex, so that the correspondence is obvious).

netaddr.go
```diff
@@ -187,9 +149,7 @@ func ParseIP(s string) (IP, error) {
 			return IPv4(ip4[0], ip4[1], ip4[2], ip4[3]), nil
 		}
 	}
-	var a16 [16]byte
-	copy(a16[:], ipa.IP.To16())
-	return IPv6Raw(a16).WithZone(ipa.Zone), nil
+	return ipv6Slice(ipa.IP.To16()).WithZone(ipa.Zone), nil
```
josharian, line 152: withInternedZone, found in https://github.com/inetaf/netaddr/pull/65, will make this more efficient.

netaddr.go
```diff
@@ -426,20 +389,24 @@ func (ip IP) IsMulticast() bool {
...
 // 128 for an IPv6 address, an error is returned.
 func (ip IP) Prefix(bits uint8) (IPPrefix, error) {
-	if ip.z == z0 {
+	effectiveBits := bits
+	switch ip.z {
+	case z0:
 		return IPPrefix{}, nil
+	case z4:
+		if bits > 32 {
+			return IPPrefix{}, fmt.Errorf("prefix length %d too large for IP address family", bits)
```
josharian, line 398: Since we know it's IPv4, let's say so in the error. Ditto below.

netaddr.go
```diff
@@ -715,40 +688,33 @@ func (p IPPrefix) IPNet() *net.IPNet {
 	}
 }
 
+func mask4(n uint8) uint32 {
+	return ^uint32(0) << (32-n)
+}
+
+func mask6(n uint8) (uint64, uint64) {
+	if n > 64 {
+		return ^uint64(0), ^uint64(0) << (64-(n-64))
```
josharian, line 697: Why `(64-(n-64))` instead of `(128-n)`?

josharian (commented):
Took a cursory first pass over this. If @bradfitz is OK with the general idea of moving to uint64s (I like it), then I'll do a second, more careful review.

Also, needs gofmt.

josharian:
Fuzz testing found a bug. See https://github.com/inetaf/netaddr/pull/66.

Time for me to get the fuzz function committed so that you can use it too. PR coming shortly.

bradfitz (commented):
I can't review this on mobile. Will check in the evening.

netaddr.go
danderson: This isn't the problematic one, that's easy to address as you describe.

The hard one is the `mask6` helper function, which constructs the masks for prefix computation. If you use little-endian and end up with swapped around bytes in the uint64, the bits are no longer contiguous, and so you can't construct the mask with simple shifts. The closest I could get was to construct the mask with shifts and then insert a bswapq _there_, rather than at IP construction time. IMO that ends up being worse.

And also, I expect x86 to have optimized the stuffing out of "mov + bswapq", given how big-endian dominates networking, so I'm not particularly bothered. There are other wins we can extract from netaddr before we're worried about byte-swapping instructions in the compiled output :)

netaddr.go
danderson: Done.

netaddr.go
danderson: I don't follow how. In the current implementation of ParseIP, we're relying on the stdlib to give us the zone as a string. We have to intern it ourselves here, we don't yet have a *intern.Value. Am I missing something? (Other than the fact that ParseIP shouldn't be shelling out to net.ParseIP, but that's a known TODO)

netaddr.go
danderson: Done.

netaddr.go
danderson: Because math at 2am is hard, apparently. Fixed.

danderson:
Fixed your feedback and the failure found by fuzz testing (the test for Is4in6 was incorrect).

netaddr.go
josharian: Oh, sorry, I got the type of `ipa` wrong. Pay me no mind. :)

josharian:
I'm wrapping up for the day and week. Looking forward to reviewing next week. This is lovely.

danderson:
For reference, https://github.com/golang/go/issues/41663 is the reason that I didn't get the same performance improvement when I kept [16]byte and did a BigEndian read-twiddle-write where needed. You might expect the resulting code to be the same, but it breaks down into byte-wise twiddling.

Whereas casting the bytes into uint64 from the get-go compiles efficiently in both sites: the BigEndian read/write are a MOVQ+BSWAPQ, and the twiddling ops are 64-bit register math.

mdlayher (approved):
LGTM, and it's a red diff! Beautiful.

netaddr.go
```diff
@@ -411,10 +385,10 @@ func (ip IP) IsLinkLocalUnicast() bool {
 // it will return false.
 func (ip IP) IsLoopback() bool {
 	if ip.Is4() {
-		return ip.a[o4+0] == 127
+		return uint8(ip.lo>>24) == 127
```
bradfitz, line 388: I'd like to avoid a number of these magic constants like "24".

How about some accessors to get at the bytes of an IPv4 address? I think they should inline away to nothing. Either one but really four might be more readable at call sites:

```go
func (ip *IP) v4a() uint8 { return uint8(ip.lo >> 24) }
func (ip *IP) v4b() uint8 { return uint8(ip.lo >> 16) }
func (ip *IP) v4c() uint8 { return uint8(ip.lo >> 8) }
func (ip *IP) v4d() uint8 { return uint8(ip.lo) }
```

(That's assuming the pointer receiver inlines better; it might not matter in which case value receiver is more proper)

netaddr.go
```diff
@@ -423,10 +397,10 @@ func (ip IP) IsLoopback() bool {
 // value, it will return false.
 func (ip IP) IsMulticast() bool {
 	if ip.Is4() {
-		return ip.a[o4+0]&0xf0 == 0xe0
+		return uint8(ip.lo>>24)&0xf0 == 0xe0
```
bradfitz, line 400: then this could be `return ip.v4a()&0xf0 == 0xe0`

netaddr.go
```diff
@@ -423,10 +397,10 @@ func (ip IP) IsLoopback() bool {
...
 	}
 	if ip.Is6() {
-		return ip.a[0] == 0xff
+		return uint8(ip.hi>>56) == 0xff
```
bradfitz, line 403: Ideally `return ip.v6(0) == 0xff` if inliner cooperates.

netaddr.go
```diff
@@ -489,7 +472,7 @@ func (ip IP) String() string {
 		return "invalid IP"
 	}
 	if ip.Is4() {
-		return fmt.Sprintf("%d.%d.%d.%d", ip.a[o4+0], ip.a[o4+1], ip.a[o4+2], ip.a[o4+3])
+		return fmt.Sprintf("%d.%d.%d.%d", uint8(ip.lo>>24), uint8(ip.lo>>16), uint8(ip.lo>>8), uint8(ip.lo))
```
bradfitz, line 475: `return fmt.Sprintf("%d.%d.%d.%d", ip.v4a(), ip.v4b(), ip.v4c(), ip.v4d())`

netaddr.go
bradfitz: Otherwise I'd also be fine with a method per byte we care about, just so it reads nicer at the call site here. e.g. `return ip.v6_0() == 0xff`

netaddr.go
```diff
@@ -291,11 +253,11 @@ func (ip IP) BitLen() uint8 {
 
 // Zone returns ip's IPv6 scoped addressing zone, if any.
 func (ip IP) Zone() string {
-	if ip.z == nil {
+	switch ip.z {
```
bradfitz, line 256: Revert this part? It's unrelated to this change and is just a stylistic preference change as far as I can tell. Can discuss on its own, but not worth mixing it with this change.

netaddr.go
```diff
@@ -399,10 +373,10 @@ func (ip IP) WithZone(zone string) IP {
 // If ip is the zero value, it will return false.
 func (ip IP) IsLinkLocalUnicast() bool {
 	if ip.Is4() {
-		return ip.a[o4+0] == 169 && ip.a[o4+1] == 254
+		return uint16(ip.lo>>16) == 0xa9fe
 	}
 	if ip.Is6() {
-		return ip.a[0] == 0xfe && ip.a[1] == 0x80
+		return uint16(ip.hi>>48) == 0xfe80
```
bradfitz, line 379: If we do octet accessor methods (assuming they inline fine), I'd use it for both bytes here separately, even if it's a bit slower. I don't think this'll be a hot spot and I'd prefer to not have a `uint16(..>> 48)`

netaddr.go
```diff
@@ -47,7 +41,22 @@ const o4 = len(mapped4Prefix)
...
+	//
+	// hi and lo are constructed by interpreting a 16-byte IPv6
+	// address as a big-endian 128-bit number. The most significant
+	// bits of that number go into hi, the rest into lo.
+	//
+	// For example, 0011:2233:4455:6677:8899:aabb:ccdd:eeff is stored as:
+	//  hi = 0x0011223344556677
+	//  lo = 0x8899aabbccddeeff
+	//
+	// We store IPs like this, rather than as [16]byte, because it
```
bradfitz, line 55: This change is okay, but the premise makes me a little sad.

How much of the 80% improvement do you get by just using the new Prefix code but keeping the old representation and doing the BigEndian.Uint64([:8]), BigEndian.Uint64([8:]) in the Prefix method first? Since the compiler *intrinsifies* those calls (_better_ than inlining), it should be pretty good.

netaddr.go
```diff
@@ -67,84 +76,48 @@ var (
...
-	return IPPrefix{IPv6Raw(ip), bits}, nil
-}
-
 // IPv6LinkLocalAllNodes returns the IPv6 link-local all nodes multicast
 // address ff02::1.
 func IPv6LinkLocalAllNodes() IP { return IPv6Raw([16]byte{0: 0xff, 1: 0x02, 15: 0x01}) }
 
 // IPv6Unspecified returns the IPv6 unspecified address ::.
-func IPv6Unspecified() IP { return IPv6Raw([16]byte{}) }
+func IPv6Unspecified() IP { return IP{z: z6noz} }
```
bradfitz, line 84: I was trying to avoid using the `z*` constants whenever possible if there was a better way to do it. I'd prefer the old way unless there's some reason. I'd hope the performance is the same?

netaddr.go
danderson: It wasn't stylistic, it was an attempt at avoiding a type assertion in the common case where the IP has no zone.

That said, looking at the SSA, the compiler optimized the prior implementation much better than I expected, so there's no benefit to doing this. Reverted.

netaddr.go
```diff
@@ -399,10 +373,10 @@ func (ip IP) WithZone(zone string) IP {
 // If ip is the zero value, it will return false.
 func (ip IP) IsLinkLocalUnicast() bool {
 	if ip.Is4() {
-		return ip.a[o4+0] == 169 && ip.a[o4+1] == 254
+		return uint16(ip.lo>>16) == 0xa9fe
```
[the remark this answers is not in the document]
danderson: Done.

netaddr.go
danderson: Based on the lack of operation coalescing I saw when doing BigEndian stuff, I expect this will degrade back to long sequences of bytewise operations, which is a pity given that we've gone to all this trouble to make the representation avoid that.

Unfortunately I can't verify what the compiler does, because `GOSSAFUNC` on any method that uses v4() or v6() seems to crash the compiler, hilariously.

netaddr.go
danderson: My personal preference is for `v4(n uint8)` and `v6(n uint8)`, for consistency. I changed the various sites to use those, except for `Is4in6` where we're doing a 48-bit comparison, and it feels silly to break that out into 6 bytewise compares.

Per above, unfortunately I can't easily inspect what the compiler produces, because GOSSAFUNC triggers a nil dereference panic when I try to analyze any function that includes calls to v4/v6.

danderson:
Managed to dump SSA, and as expected the compiler inlines, but doesn't do any intelligent coalescing. The check `v4(0) == 169 && v4(1) == 254` ends up doing:

```
MOVQ "".ip+8(SP), AX  # Load all of ip.lo into A
MOVQ AX, CX # Copy A to C
SHRQ $24, AX # Shift A right 24
CMPB AX, $-87 # v4(0) == 169
JNE 19
SHRQ $16, CX # Shift C right 16
CMPB CX, $-2 # v4(1) == 254
```

Compare to `uint16(ip.lo>>16) == 0xa9fe`:

```
MOVQ "".ip+8(SP), AX # Load all of ip.lo into A
SHRQ $16, AX # Shift A right 16
CMPW AX, $-22018 # Compare lower 16 bits of register to 0xa9fe
```

So, the generated code is definitely worse, and in line with what I saw when I tried to use encoding/binary to Uint64+twiddle+PutUint64. Based on that, I would prefer to continue operating on larger chunks of the address, rather than break it down bytewise. Maybe something like `v4ab() uint16`, and similar custom accessors for each different slice of IP that we need to examine? That would still make the high-level functions more readable, at the expense of a bit more duplication of bit twiddlers.

netaddr.go
josharian: They’re not intrinsics. That’s part of how we ended up with broken compiler optimizations for them. See also https://github.com/golang/go/issues/42958, in which I suggest making them incrementally more intrinsic-like.

netaddr.go
josharian: If you can crash the compiler using GOSSAFUNC, please file an upstream issue! Those can be leading indicators of other bugs.

BTW, for the future you can get assembly out with -gcflags=-S, although that’ll be for the whole package, and it’ll get dumped to stdout.

danderson: David Chase looks to be on it already, otherwise would have filed, yes.

netaddr.go
josharian: Also worth noting is that the compiler can (theoretically) always do cheap binary tricks on amd64, which allows unaligned access. But without a way to specify alignment (such as https://github.com/golang/go/issues/19057 or https://github.com/golang/go/issues/6752), the compiler can’t be sure that the struct is aligned, arm systems can only benefit when we use uint64s.

bradfitz: Cool, thanks for the links. Can you link all those (or at least link to this PR thread) in the commit message?

By default it seems difficult to map back from a commit back to its code review on GitHub.

netaddr.go
bradfitz: Not strongly. It's fine as you have it.

bradfitz:
> Based on that, I would prefer to continue operating on larger chunks of the address, rather than break it down bytewise. 

I'd prefer to err on the side of explicit & readable code until there's a performance reason to micro-optimize `IsLinkLocalUnicast`.

Out of curiosity, does Go do any better at `ip.v4(0)<<8|ip.v4(1) == (169<<8)|254` ?

netaddr.go
danderson: Added links to this PR, and to the Go bugs we've linked from here, along with an explanatory paragraph explaining "why not just change Prefix?"

danderson:
Not by much. The compare becomes at 16-byte CMPW, but the load is still dumb (loads ip.lo twice into two different registers, shifts each one by a different amount, then ORs them together).

Keeping at bytewise loads for now. Will see in a followup if I can make it generate slightly less rubbish code without compromising readability.

bradfitz:
LGTM at b7d6c09. Thanks! Merge away.

PR #66, opened 2020-12-24
add ParseIP test case

josharian:
This passes on master but fails in #63 (at the moment).
This makes it interesting enough to keep.


mdlayher (approved):
LGTM

josharian:
Oops. This made a merge commit. :(((( I just changed the settings to disallow that.

PR #67, opened 2020-12-24
add a fuzz function and corpus

josharian:
I decided in this case to also commit the corpus, because it has been built up over several different implementations of netaddr, which means it has coverage hints that might be hard to discover given only one of those implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,17 @@ See its docs: https://godoc.org/inet.af/netaddr
 * merges net.IPAddr and net.IP (which the Go net package is a little torn between for legacy reasons)
 * ...
 * TODO: finish this list
+
+## Testing
+
+In addition to regular Go tests, netaddr uses fuzzing.
+The corpus is stored separately, in a submodule,
+to minimize the impact on everyone else.
+
+To use:
+
+```
+$ git submodule update --init
+$ go get -u github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build
+$ go-fuzz-build && go-fuzz
+```
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+crashers
+suppressions
+netaddr-fuzz.zip
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "corpus"]
+	path = corpus
+	url = git@github.com:inetaf/netaddr-corpus.git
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -18,3 +18,14 @@ See its docs: https://godoc.org/inet.af/netaddr
...
 * TODO: finish this list
+
+## Testing
+
+In addition to regular Go tests, netaddr uses fuzzing.
+To use:
+
+```
+$ go get -u github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build
+$ go-fuzz-build && go-fuzz
```
mdlayher, line 29: I feel like I've had to use `CGO_ENABLED=0 go-fuzz-build` for a long time on linux/amd64 to get things working. Have you run into that?

mdlayher (approved):
LGTM

README.md
josharian: Oh, I see https://github.com/dvyukov/go-fuzz/issues/101. I've not run into it. 2016 is a long time ago, and I re-worked large sections of go-fuzz-build to use go/packages. I wonder whether maybe this is fixed now. (Fingers crossed!)

josharian:
Thanks for the review. I'm going to wait on a review from Brad, because I think he may have strong feelings about checking in the corpus.

mdlayher:
No worries. I made a separate fuzz corpus repo for my projects just to keep my code repos tidy, but I don't have strong feelings either way.

danderson:
Adding corpus SGTM. If we want to avoid the noise of generated files in the main repo, we could do a sub-repo for the fuzzing corpus?

bradfitz:
Can we do the corpus in a separate branch (of the same repo) instead? Then the "make fuzz" target or whatever can "git fetch" it as needed (at least if empty). I did that with storing the linux kernel for integration tests before (https://docs.google.com/presentation/d/1rAAyOTCsB8GLbMgI0CAbn69r6EVWL8j3DPl4qc0sSlc/edit#slide=id.g410de087f7_1_31) which was nice... reduced noise & download for regular users.

josharian:
The corpus is going to be more read-write than the kernel, and it's a directory full of files. I'm more inclined to go the submodule route, if that works for you. I've created a corpus submodule and updated this PR for your perusal. WDYT?

mdlayher (approved):
LGTM with submodule but will leave for @bradfitz.

bradfitz (approved):
submodule SGTM

PR #69, opened 2020-12-24
add IPPrefix.IsSingleIP

josharian:
I just encountered the second place where this was implemented in Tailscale, each different ways.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -623,6 +623,9 @@ type IPPrefix struct {
 // IsZero reports whether p is its zero value.
 func (p IPPrefix) IsZero() bool { return p == IPPrefix{} }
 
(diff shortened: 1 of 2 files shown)


mdlayher (approved):
LGTM. This is something I've rolled by hand as well.

bradfitz (commented):
Lgtm

danderson:
Once again I wish Go had uint128, because Prefix could really use a Size() as well, which would subsume this. But being unable to represent 2^128 sanely, this LGTM.

josharian:
Please add a short note to https://github.com/golang/go/issues/9455. Use cases like this can help drive such proposals forward. I want uint128 too.

PR #70, opened 2020-12-24
netaddr: add MustParse* helpers

josharian:
Several Tailscale packages each have their own copy of these helpers.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -192,6 +192,16 @@ func ParseIP(s string) (IP, error) {
 	return IPv6Raw(a16).WithZone(ipa.Zone), nil
 }
 
+// MustParseIP calls ParseIP(s) and panics on error.
+// It is intended for use in tests with hard-coded strings.
+func MustParseIP(s string) IP {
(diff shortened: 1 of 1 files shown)


bradfitz (approved):
I'm lukewarm on this. I guess regexp and template have them. They seem to invite misuse but whatever. They're named scarily and conventionally at least. 

How do others feel?

mdlayher (approved):
I add these in my test files all the time, so I'd be +1.

danderson:
I've written these so many times, I really want them somewhere, if not here then in a tailscale util package.

Misuse is a possibility, but the Must* convention is well understood in Go, so you get what you pay for.

Minor bikeshed: `MustIP` and `MustIPPrefix` ? That's what I all these in test code.

bradfitz:
Let's not try to shorten them. Test code likely will do that anyway to remove the "netaddr." prefix anyway.

danderson:
Fair 'nuf.

josharian:
Sounds like consensus enough. Merging.

PR #74, opened 2020-12-26
Add a slow, but correct ParseIP as a reference implementation.

danderson:
The goal of this implementation is to be slow, but very readable, and a point of comparison for optimized implementations: faster implementations should behave exactly like parseIPSlow on the set of valid and invalid inputs, and the tests now enforce that in addition to parity with net.ParseIP.

Mostly this is useful if we implement behavior that diverges from net.ParseIP. By having a reference parser in netaddr itself, we can evolve the optimized/not-as-readable implementation in tandem with our own reference impl.

### DIFF
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -17,186 +17,278 @@ import (
 	"sort"
 	"strings"
 	"testing"
+
+	"go4.org/intern"
 )
 
-func TestParseString(t *testing.T) {
-	tests := []string{
-		"1.2.3.4",
-		"0.0.0.0",
-		"::",
-		"::1",
-		"::1%zone",
-		"fe80::1cc0:3e8c:119f:c2e1%ens18",
-		"::ffff:c000:1234",
-		"::ffff:f077:ff",
-		"::ffff:c000:1234%zone",
-		"::f:ffff:0:0",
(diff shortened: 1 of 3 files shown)


netaddr_test.go
```diff
@@ -17,186 +17,268 @@ import (
...
 			if err != nil {
 				t.Fatal(err)
 			}
-			ip2, err := ParseIP(s)
+			if got != test.ip {
+				t.Errorf("ParseIP(%q) got %#v, want %#v", test.in, got, test.ip)
+			}
+
+			// Check that parsing is stable.
+			got2, err := ParseIP(test.in)
```
mdlayher, line 120: Stable as in, there is no intermediate state established in the underlying parser that results in a different result when parsing the same input string again? I would have expected something like `got.String()` as an input to test round tripping.

mdlayher (approved):
LGTM.

netaddr_test.go
danderson: Yeah, I guess strictly speaking this is a test that ParseIP is a pure function.

danderson: Added a test that `ParseIP(ip.String()) == ip`.

PR #75, opened 2020-12-26
Speed up parseIPv4 3.3x.

danderson:
```
name            old time/op    new time/op    delta
ParseIPv4-8       63.2ns ± 4%    20.1ns ± 4%  -68.17%  (p=0.000 n=19+19)
StdParseIPv4-8    93.9ns ± 3%    93.3ns ± 3%     ~     (p=0.211 n=20+19)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -179,24 +179,51 @@ func MustParseIP(s string) IP {
 // net.ParseIP at the moment, so returning ok=false from here just
 // means it'll try it a slower way.
 func parseIPv4(s string) (ip IP, ok bool) {
-	if len(s) < 7 || len(s) > 15 || strings.Count(s, ".") != 3 {
-		return ip, false
-	}
-	var v [4]uint64
-	for i := range v {
-		var numStr string
-		if dot := strings.IndexByte(s, '.'); dot == -1 {
-			numStr = s
-		} else {
-			numStr, s = s[:dot], s[dot+1:]
+	var ip4 [4]byte
+
+	for i := 0; i < 4; i++ {
+		var (
+			j   int
+			acc uint16
+		)
+		// Parse one byte of digits. Bail if we overflow, stop at
+		// first non-digit.
+		//
+		// As of Go 1.15, don't try to factor this digit reading into
+		// a helper function. Its complexity is slightly too high for
+		// inlining, which ends up costing +50% in parse time.
+		for j = 0; j < len(s); j++ {
+			if s[j] < '0' || s[j] > '9' {
+				break
+			}
+			acc = (acc * 10) + uint16(s[j]-'0')
+			if acc > 255 {
+				return IP{}, false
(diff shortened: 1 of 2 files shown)


netaddr.go
```diff
@@ -179,24 +179,55 @@ func MustParseIP(s string) IP {
...
-		return ip, false
-	}
-	var v [4]uint64
-	for i := range v {
-		var numStr string
-		if dot := strings.IndexByte(s, '.'); dot == -1 {
-			numStr = s
-		} else {
-			numStr, s = s[:dot], s[dot+1:]
+	// Parse directly into the return value. This shaves another ~5%
```
bradfitz, line 182: I'd rather pay that 5% cost.

netaddr.go
```diff
@@ -179,24 +179,55 @@ func MustParseIP(s string) IP {
...
-		var err error
-		v[i], err = strconv.ParseUint(numStr, 10, 8)
-		if err != nil {
-			return ip, false
+		// There must be at least 1 digit per quad.
+		if j == 0 {
+			return IP{}, false
+		}
+
+		ip.lo = (ip.lo<<8)+acc
```
bradfitz, line 213: or `|` instead of `+` but doubt it matters

netaddr.go
mdlayher: I assume you'd prefer that the resulting IP always be zero if !ok?

netaddr.go
```diff
@@ -179,24 +179,55 @@ func MustParseIP(s string) IP {
...
+
+		// Non-final byte must be followed by a dot
+		if i < 3 {
+			if len(s) == j || s[j] != '.' {
```
kf6nux, line 217: It looks `len(s) == j ` is meant to catch missing quads, but the current missing quad tests (Class A/B) are caught by `if acc > 255`. Might want to add a test case of `127.1` being invalid?

netaddr.go
danderson: I dislike this focus on always using the base constructors even when they make the computer's job harder, but under duress, done. Benchmark is now 20ns/IP instead of 19ns/IP.

netaddr.go
danderson: I tried both while implementing, no change in performance, so I kept the one that felt more obvious. That said, this code is gone now that the function uses IPv4().

netaddr.go
danderson: Added test cases for 127.1 and 127.0.1. Thanks for catching that this branch was missing coverage!

netaddr.go
danderson: @mdlayher my preference is to always return a zero type on failure, because otherwise we open ourselves up to someone not checking the error and ending up using a half-built IP address.

It's moot with the change back to IPv4() though, since the return value no longer has intermediate state in it.

PR #76, opened 2020-12-27
Parse IPv6 without net.ParseIP, and make it faster.

danderson:
With this change, netaddr is no longer dependent on net.ParseIP for its
    ParseIP implementation. It's also 2-4x faster than net.ParseIP and
    allocation-free, except for v6+zone addresses which are still slower and
    require 1 allocation.

```    
    name                      old time/op    new time/op    delta
    ParseIP/v4-8                22.0ns ± 2%    24.5ns ± 1%   +11.46%  (p=0.016 n=5+4)
    ParseIP/v6-8                 169ns ± 2%      91ns ± 2%   -45.97%  (p=0.008 n=5+5)
    ParseIP/v6_ellipsis-8        140ns ± 2%      68ns ± 5%   -51.32%  (p=0.008 n=5+5)
    ParseIP/v6_v4-8              216ns ± 1%      71ns ± 1%   -67.19%  (p=0.008 n=5+5)
    ParseIP/v6_zone-8            526ns ± 1%     222ns ± 1%   -57.89%  (p=0.008 n=5+5)
    StdParseIP/v4-8             94.5ns ± 2%    89.0ns ± 1%    -5.78%  (p=0.008 n=5+5)
    StdParseIP/v6-8              150ns ± 1%     158ns ± 2%    +5.21%  (p=0.008 n=5+5)
    StdParseIP/v6_ellipsis-8     116ns ± 3%     125ns ± 2%    +7.59%  (p=0.008 n=5+5)
    StdParseIP/v6_v4-8           190ns ± 1%     183ns ± 2%    -3.89%  (p=0.008 n=5+5)
    StdParseIP/v6_zone-8         138ns ± 2%     133ns ± 2%    -3.20%  (p=0.024 n=5+5)
    
    name                      old alloc/op   new alloc/op   delta
    ParseIP/v4-8                 0.00B          0.00B           ~     (all equal)
    ParseIP/v6-8                 16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_ellipsis-8        16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_v4-8              32.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_zone-8            64.0B ± 0%     16.0B ± 0%   -75.00%  (p=0.008 n=5+5)
    StdParseIP/v4-8              16.0B ± 0%     16.0B ± 0%      ~     (all equal)
    StdParseIP/v6-8              16.0B ± 0%     16.0B ± 0%      ~     (all equal)
    StdParseIP/v6_ellipsis-8     16.0B ± 0%     16.0B ± 0%      ~     (all equal)
    StdParseIP/v6_v4-8           32.0B ± 0%     32.0B ± 0%      ~     (all equal)
    StdParseIP/v6_zone-8         16.0B ± 0%     16.0B ± 0%      ~     (all equal)
    
    name                      old allocs/op  new allocs/op  delta
    ParseIP/v4-8                  0.00           0.00           ~     (all equal)
    ParseIP/v6-8                  1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_ellipsis-8         1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_v4-8               2.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
    ParseIP/v6_zone-8             4.00 ± 0%      1.00 ± 0%   -75.00%  (p=0.008 n=5+5)
    StdParseIP/v4-8               1.00 ± 0%      1.00 ± 0%      ~     (all equal)
    StdParseIP/v6-8               1.00 ± 0%      1.00 ± 0%      ~     (all equal)
    StdParseIP/v6_ellipsis-8      1.00 ± 0%      1.00 ± 0%      ~     (all equal)
    StdParseIP/v6_v4-8            2.00 ± 0%      2.00 ± 0%      ~     (all equal)
    StdParseIP/v6_zone-8          1.00 ± 0%      1.00 ± 0%      ~     (all equal)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -124,43 +124,19 @@ func IPFrom16(addr [16]byte) IP {
 // s can be in dotted decimal ("192.0.2.1"), IPv6 ("2001:db8::68"),
 // or IPv6 with a scoped addressing zone ("fe80::1cc0:3e8c:119f:c2e1%ens18").
 func ParseIP(s string) (IP, error) {
-	// IPv4 fast path.
-	if ip, ok := parseIPv4(s); ok {
-		return ip, nil
-	}
-
-	var ipa net.IPAddr
-	ipa.IP = net.ParseIP(s)
-	if ipa.IP == nil {
-		switch percent := strings.Index(s, "%"); percent {
-		case -1:
-			// handle bad input with no % at all, so the net.ParseIP was not due to a zoned IPv6 fail
-			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse IP", s)
-		case 0:
-			// handle bad input with % at the start
-			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing IPv6 address", s)
-		case len(s) - 1:
-			// handle bad input with % at the end
-			return IP{}, fmt.Errorf("netaddr.ParseIP(%q): missing zone", s)
-		default:
-			// net.ParseIP can't deal with zoned scopes, let's split and try to parse the IP again
-			s, ipa.Zone = s[:percent], s[percent+1:]
-			ipa.IP = net.ParseIP(s)
-			if ipa.IP == nil {
-				return IP{}, fmt.Errorf("netaddr.ParseIP(%q): unable to parse IP", s)
-			}
+	for i := 0; i < len(s); i++ {
+		switch s[i] {
+		case '.':
+			return parseIPv4(s)
+		case ':':
+			return parseIPv6(s)
+		case '%':
+			// Assume that this was trying to be an IPv6 address with
+			// a zone specifier, but the address is missing.
+			return IP{}, parseIPErrorf(s, "missing IPv6 address")
 		}
 	}
-
-	if !strings.Contains(s, ":") {
-		if ip4 := ipa.IP.To4(); ip4 != nil {
-			if ipa.Zone != "" {
-				return IP{}, fmt.Errorf("netaddr.ParseIP(%q): invalid zone with IPv4 address", s)
-			}
-			return IPv4(ip4[0], ip4[1], ip4[2], ip4[3]), nil
-		}
-	}
-	return ipv6Slice(ipa.IP.To16()).WithZone(ipa.Zone), nil
+	return IP{}, errors.New("unable to parse IP")
 }
 
 // MustParseIP calls ParseIP(s) and panics on error.
@@ -173,12 +149,14 @@ func MustParseIP(s string) IP {
 	return ip
 }
 
-// parseIPv4 parses s as an IPv4 address (in form "192.168.0.1") and
-// reports whether it parsed as such. This is a fast path for ParseIP
-// for now. It ultimately falls back to the standard library's
-// net.ParseIP at the moment, so returning ok=false from here just
-// means it'll try it a slower way.
-func parseIPv4(s string) (ip IP, ok bool) {
+func parseIPErrorf(ip, msg string, args ...interface{}) error {
+	msg = "ParseIP(%q): " + msg
+	args = append([]interface{}{ip}, args...)
+	return fmt.Errorf(msg, args...)
+}
+
+// parseIPv4 parses s as an IPv4 address (in form "192.168.0.1").
+func parseIPv4(s string) (ip IP, err error) {
 	var ip4 [4]byte
 
 	for i := 0; i < 4; i++ {
@@ -198,20 +176,22 @@ func parseIPv4(s string) (ip IP, ok bool) {
 			}
 			acc = (acc * 10) + uint16(s[j]-'0')
 			if acc > 255 {
-				return IP{}, false
+				return IP{}, parseIPErrorf(s, "IPv4 field has value >255")
 			}
 		}
 		// There must be at least 1 digit per quad.
 		if j == 0 {
-			return IP{}, false
+			return IP{}, parseIPErrorf(s, "each dot-separated field must have at least one digit")
 		}
 
 		ip4[i] = uint8(acc)
 
 		// Non-final byte must be followed by a dot
 		if i < 3 {
-			if len(s) == j || s[j] != '.' {
-				return IP{}, false
+			if len(s) == j {
+				return IP{}, parseIPErrorf(s, "address too short")
+			} else if s[j] != '.' {
+				return IP{}, parseIPErrorf(s, "unexpected character %q", s[j])
 			}
 			j++
 		}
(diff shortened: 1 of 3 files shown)


danderson:
Still a draft because I need to fix up the error handling and comments, but opening the PR early for a question: after a bunch of false starts implementing a fancy parseIPv6, I took the implementation from the Go net package and adapted it. It's changed a fair bit, but the basic structure remains the same. How should we handle the licensing for that? Put it in a separate file and acknowledge the Go license for that?

netaddr_test.go
```diff
@@ -1441,31 +1443,38 @@ func BenchmarkIPv4Contains(b *testing.B) {
...
-}
-
-func BenchmarkStdParseIPv4(b *testing.B) {
-	b.ReportAllocs()
-	for i := 0; i < b.N; i++ {
-		net.ParseIP("192.168.1.1")
+func BenchmarkParseIP(b *testing.B) {
+	z := intern.Get("eth1") // Pin to not benchmark the intern package
+	_ = z
+	for _, test := range parseBenchInputs {
```
mdlayher, line 1460: Any chance you'd want to reset the allocations count once the intern pool is set up?

danderson: No change. Each case runs as a sub-benchmark, which I think implicitly resets the counters.

I'm still not sure where the extra alloc comes from in the v6+zone path. Either I'm doing the interning wrong and intern.Get allocates every time, or it's some unavoidable allocation in the intern package (maybe boxing the string in an interface{}?).

danderson:
Ready for review. Parsers return nice errors now, and are still fast (except the v6+zone case, I still don't know how it manages to be 2x slower than stdlib)

netaddr.go
```diff
@@ -173,12 +149,14 @@ func MustParseIP(s string) IP {
 	return ip
 }
 
-// parseIPv4 parses s as an IPv4 address (in form "192.168.0.1") and
-// reports whether it parsed as such. This is a fast path for ParseIP
-// for now. It ultimately falls back to the standard library's
-// net.ParseIP at the moment, so returning ok=false from here just
-// means it'll try it a slower way.
-func parseIPv4(s string) (ip IP, ok bool) {
+func parseErr(ip, msg string, args ...interface{}) error {
```
josharian, line 152: Call this `parseIPErr`, because we have other parsers.

I saw a tweet from you about how adding this made things slower. Fanciful theory: It's pushing stuff out of the instruction cache. Does it help if you use a bespoke error type and do the formatting work later?

```go
type ipParseError struct {
  input string
  msg string
  args ...interface{}
}

func parseIPErr(ip, msg string args ...interface{} error {
  return &ipParseError{input, msg, args}
}

func (e *ipParseError) Error() string {
  args := append([]interface{}{e.input}, e.args...)
  return fmt.Sprintf("ParseIP(%q): "+e.msg, args...)
}
```

netaddr.go
```diff
@@ -220,10 +200,152 @@ func parseIPv4(s string) (ip IP, ok bool) {
...
+			} else if c >= 'a' && c <= 'f' {
+				acc = (acc << 4) + uint32(c-'a'+10)
+			} else if c >= 'A' && c <= 'F' {
+				acc = (acc << 4) + uint32(c-'A'+10)
+			} else {
+				break
+			}
+			if acc > maxUint16 {
+				// Overflow, fail.
+				return IP{}, parseErr(s, "IPv6 field has value >2^16")
```
josharian, line 261: `>=`?

netaddr_test.go
```diff
@@ -1441,31 +1445,38 @@ func BenchmarkIPv4Contains(b *testing.B) {
...
+	z := intern.Get("eth1") // Pin to not benchmark the intern package
+	for _, test := range parseBenchInputs {
+		b.Run(test.name, func(b *testing.B) {
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
+				sinkIP, _ = ParseIP(test.ip)
+			}
+		})
 	}
+	_ = z // keep the interned value alive during sub-benchmarks
```
josharian, line 1469: This will get optimized away. Use a global sink instead. That might explain the allocation. If it doesn't explain the allocation, run with `-memprofile` to find it. (I can help with that if needed.)

josharian (commented):
Haven't given this a thorough look, but wanted to get in a few small comments.

netaddr.go
mdlayher: A nit: anything that returns a formatted string/error should be named as such, maybe "parseIPErrorf" or just "ipErrorf"?

netaddr.go
mdlayher: I think `>` would be correct because `ffff` would be 2^16 and valid, yeah?

netaddr.go
```diff
@@ -220,10 +200,152 @@ func parseIPv4(s string) (ip IP, ok bool) {
...
+			return IP{}, parseErr(s, "zone specifier not allowed with IPv4 addresses")
+		}
+		return IP{}, parseErr(s, "trailing garbage after address")
+	}
+
+	return IPv4(ip4[0], ip4[1], ip4[2], ip4[3]), nil
+}
+
+// parseIPv6 parses s as an IPv6 address (in form "2001:db8::68").
+func parseIPv6(s string) (IP, error) {
```
mdlayher, line 213: Should add a test case for something like `2001:db8:a:b:c:d:e::` or `2001:db8::a:b:c:d:e` and weird edge cases like a single block of zeros being compressed to `::` or trailing segments like that.

netaddr_test.go
mdlayher: Ack, unfortunately I don't have any more ideas.

mdlayher (approved):
Looks awesome overall, just a few nits and questions.

netaddr.go
josharian: Another possibility is it makes the layout pass change its mind about which branch is unlikely, making the branch predictor unhappy. (The internet claims that static branch prediction isn't a thing and doesn't matter. It is empirically wrong about that.)

netaddr.go
josharian:
```python
>>> 0xffff
65535
>>> 2**16
65536
```

danderson: maxUint16 is 2^16-1, so > is correct. >= would reject the input `::ffff`, which is a valid IP.

josharian: Sorry to be unclear. I am referring to the error message string, not the comparison two lines up.

netaddr_test.go
danderson: No change. memprofile says the 16B allocation happens in `IP.WithZone`, at `ip.z = intern.Get(zone)`. That the allocation is 16B makes me suspect the conversion of zone from string to interface{}. I'm not clear on why that allocation needs to happen.

danderson: With gcflags `-m -m`:

```
./netaddr.go:758:42: parameter zone leaks to {heap} with derefs=0:
./netaddr.go:758:42:   flow: {heap} = zone:
./netaddr.go:758:42:     from ip.WithZone(zone) (call parameter) at ./netaddr.go:769:19
```

netaddr.go
danderson: oh, doh. Yes.

netaddr_test.go
danderson: Loose hypothesis: in creating the interface{} for the call to intern.Get, the compiler can't prove that the interface won't outlive the zone string, and so is forced to evict it to the heap.

In practice, if there's a cache hit within intern.Get, the interface{} won't outlive the WithZone call, but the compiler can't prove it, and so any WithZone call with a non-empty zone string will force a heap allocation for a string slice.

I don't think there's any way to avoid that, unless intern exposes a specialized GetString API (would be easier with generics :( ).

netaddr.go
danderson: Done.

netaddr_test.go
danderson: Also checked with -cpuprofile, and confirmed that v6+zone's slowness comes entirely from WithZone, about 50% spent allocating that escaping string, and the other 50% in intern.Get doing the map lookup and whatnot.

So, the conclusion is just: at the timescales we're working at here, using the intern package is quite expensive, and unless we can make it significantly cheaper, v6+zone will always be ~2x slower than the stdlib. My take is that the slowness for that edge case is worth it in return for the other nice things that interned zoneptrs get us.

netaddr_test.go
josharian: https://github.com/go4org/intern/pull/10 if we want it

danderson: With GetString, things are happier:

```
name                  old time/op    new time/op    delta
ParseIP/v6_zone-8        221ns ± 3%     164ns ± 1%   -26.04%  (p=0.008 n=5+5)
StdParseIP/v6_zone-8     132ns ± 2%     134ns ± 2%      ~     (p=0.056 n=5+5)

name                  old alloc/op   new alloc/op   delta
ParseIP/v6_zone-8        16.0B ± 0%      0.0B       -100.00%  (p=0.008 n=5+5)
StdParseIP/v6_zone-8     16.0B ± 0%     16.0B ± 0%      ~     (all equal)

name                  old allocs/op  new allocs/op  delta
ParseIP/v6_zone-8         1.00 ± 0%      0.00       -100.00%  (p=0.008 n=5+5)
StdParseIP/v6_zone-8      1.00 ± 0%      1.00 ± 0%      ~     (all equal)
```

Still slightly slower than stdlib on v6+zone, but only very slightly, and because we're doing justified extra work to get nice value types.

danderson: It's a nicely separable change and needs to wait until intern is updated, so I'm going to defer this until the next PR. This PR is still a net improvement in every dimension without the intern.GetString change.

netaddr.go
danderson: Merging this PR before I dig deeper into the error things (other than the name change, which I've done). The PR is already quite large and even with the mild regression introduced by error handling, it's much better than what's at HEAD right now.

PR #78, opened 2020-12-28
netaddr: add gcassert annotations to ensure function inlining

### LINKED ISSUE #77: Make use of gcassert utility to assert that code is inlined and/or BCE'd
This tool is quite useful and can be used to annotate code to ensure the compiler is doing what we'd like.

https://github.com/jordanlewis/gcassert

I think this is probably worth adding to CI as well.

/cc @jordanlewis

mdlayher:

I mostly used some hand-waving to figure out functions that looked simple or critical enough to merit the annotation, but I'm sure there are more to be done. If this is not something we'd like to maintain, that's okay, but I figured I would open this for discussion.

Fixes #77.

### DIFF
--- a/.github/workflows/gcassert.yml
+++ b/.github/workflows/gcassert.yml
@@ -1,0 +1,25 @@
+name: gcassert
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.15
+
+    - name: Check out code
+      uses: actions/checkout@v1
+
+    - name: Run gcassert
+      run: go run github.com/jordanlewis/gcassert/cmd/gcassert
--- a/netaddr.go
+++ b/netaddr.go
@@ -260,12 +260,14 @@ func FromStdIPRaw(std net.IP) (ip IP, ok bool) {
 
 // v4 returns the i'th byte of ip. If ip is not an IPv4, v4 returns
 // unspecified garbage.
+//gcassert:inline
 func (ip IP) v4(i uint8) uint8 {
 	return uint8(ip.lo >> ((3 - i) * 8))
 }
 
 // v6 returns the i'th byte of ip. If ip is an IPv4 address, this
 // accesses the IPv4-mapped IPv6 address form of the IP.
+//gcassert:inline
 func (ip IP) v6(i uint8) uint8 {
 	if i >= 8 {
 		return uint8(ip.lo >> ((15 - i) * 8))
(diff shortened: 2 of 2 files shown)


netaddr.go
```diff
@@ -278,12 +280,14 @@ func (ip IP) v6(i uint8) uint8 {
 // The zero value is not a valid IP address of any type.
 //
 // Note that "0.0.0.0" and "::" are not the zero value.
+//gcassert:inline
```
bradfitz, line 283: We don't want this in the godoc. Or do all the tools strip these?

mdlayher: I had the same concern, but it seems like `go doc` at least removes it:
```
✔ ~/src/inet.af/netaddr [mdl-gcassert|⚑ 1] 
17:16 $ go doc IP.IsZero
package netaddr // import "inet.af/netaddr"

func (ip IP) IsZero() bool
    IsZero reports whether ip is the zero value of the IP type. The zero value
    is not a valid IP address of any type.

    Note that "0.0.0.0" and "::" are not the zero value.

```

bradfitz: Is that done at a low layer so godoc.org etc also gets it?

.github/workflows/gcassert.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.15
+
+    - name: Check out code
+      uses: actions/checkout@v1
+
+    - name: Run gcassert
+      run: go run github.com/jordanlewis/gcassert/cmd/gcassert
```
bradfitz, line 25: I think I'd prefer to just this part of our existing tests so "go test ./..." catches it, rather than remembering to run some separate tool.

I'd fine importing `gcassert`'s guts in our testing code, though.

The standard library has a number of tests that check for this sort of stuff (running the compiler or `go tool nm`) and I always liked just having them part of a normal test.

josharian: Seconded.

netaddr.go
mdlayher: I can confirm that `golang.org/x/tools/cmd/godoc` ignores the gcassert annotations, so presumably godoc.org and pkg.go.dev would as well.

.github/workflows/gcassert.yml
mdlayher: Sure, looks like it'll be easy enough. Can do.

mdlayher: IANAL, but here's the license header from the tool before I start importing its code:

```
// Copyright 2020 The Cockroach Authors.
//
// Use of this software is governed by the Business Source License
// included in the file licenses/BSL.txt.
//
// As of the Change Date specified in that file, in accordance with
// the Business Source License, use of this software will be governed
// by the Apache License, Version 2.0, included in the file
// licenses/APL.txt.
```

Any objections to that?

mdlayher: /cc @jordanlewis

bradfitz: Naah, let's not mix in non-open source code. I'll write it myself.

bradfitz:
We'll do this differently.

.github/workflows/gcassert.yml
jordanlewis: Argh, this doesn't need to be BSL at all. I can change the license to something unrestrictive, I'll re-ping here when I do.

jordanlewis: I updated the license to MIT. I'm guessing you won't use it anyway, but on the off chance you want to, it's MIT licensed.

bradfitz: @jordanlewis, thanks! I'll leave it to @mdlayher on whether he wants to update my https://github.com/inetaf/netaddr/commit/be7e645283661d9d64c0e6731e8df7820a9f3dc6 version with gcassert.

PR #88, opened 2020-12-29
github: add Go 1.15 to matrix, keep 1.11, 1.14, drop 1.12, 1.13

bradfitz:


danderson (approved):
By supporting 1.11, we're going well above and beyond Go's own support promises. Is it worth it?

(no strong feelings, it currently doesn't cost much to keep supporting the older versions)

bradfitz:
I don't have strong feelings about it. But it's also not hard to support older Go either, especially when it's such a low-level package. I'm happy to bump it when a real need arises.

PR #89, opened 2020-12-29
add IPRange.Prefix

bradfitz:
Allocation-free way to test whether an IPRange is an IPPrefix and get it.

(Needs tests before submitting but I gotta get off the computer now.)

Needed this for some Tailscale stuff.

### DIFF
--- a/inlining_test.go
+++ b/inlining_test.go
@@ -42,7 +42,6 @@ func TestInlining(t *testing.T) {
 	})
 	for _, want := range []string{
 		"(*IPSet).Add",
-		"(*IPSet).Ranges.func1",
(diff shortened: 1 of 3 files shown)


inlining_test.go
```diff
@@ -42,7 +42,6 @@ func TestInlining(t *testing.T) {
 	})
 	for _, want := range []string{
 		"(*IPSet).Add",
-		"(*IPSet).Ranges.func1",
```
bradfitz, line 45: I meant to delete these earlier.

bradfitz:
PTAL; now with tests.

danderson (approved):

PR #90, opened 2020-12-30
introduce parseIPError as an explicit type.

danderson:
This lets us provide slightly better parse error information (including
the position at which the error occurred), and also speeds up successful
parsing in some cases, because error formatting is delayed beyond the
execution of ParseIP.

```
name                   old time/op    new time/op    delta
ParseIP/v4-8             24.9ns ± 1%    24.0ns ± 2%  -3.58%  (p=0.000 n=8+10)
ParseIP/v6-8             90.0ns ± 4%    92.7ns ± 4%  +3.03%  (p=0.006 n=10+10)
ParseIP/v6_ellipsis-8    69.1ns ±10%    62.4ns ± 5%  -9.60%  (p=0.000 n=9+9)
ParseIP/v6_v4-8          73.4ns ± 4%    66.9ns ± 3%  -8.81%  (p=0.000 n=9+9)
ParseIP/v6_zone-8         221ns ± 2%     215ns ± 2%  -2.58%  (p=0.000 n=10+10)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -133,7 +133,7 @@ func ParseIP(s string) (IP, error) {
 		case '%':
 			// Assume that this was trying to be an IPv6 address with
 			// a zone specifier, but the address is missing.
-			return IP{}, parseIPErrorf(s, "missing IPv6 address")
+			return IP{}, parseIPError{in: s, msg: "missing IPv6 address"}
 		}
 	}
 	return IP{}, errors.New("unable to parse IP")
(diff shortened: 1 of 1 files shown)


danderson:
@josharian implementation of your earlier suggestion. The effects on performance are quite marginal (which is also why I declined to care much about the impact in the last PR), but I believe the parse errors are a bit better with this change, and it also happens to move the performance needle marginally in a happy direction.

josharian (approved):

PR #92, opened 2020-12-30
Optimizes parseIPv4

AlexanderYastrebov:
```
benchmark                              old ns/op     new ns/op     delta
BenchmarkParseIP/v4-8                  22.9          18.4          -19.65%
BenchmarkParseIP/v6-8                  77.4          77.9          +0.65%
BenchmarkParseIP/v6_ellipsis-8         56.5          55.5          -1.77%
BenchmarkParseIP/v6_v4-8               63.8          58.1          -8.93%
BenchmarkParseIP/v6_zone-8             144           140           -2.78%
BenchmarkParseIPPort/v4-8              42.5          37.8          -11.06%
BenchmarkParseIPPort/v6-8              99.6          100           +0.40%
BenchmarkParseIPPort/v6_ellipsis-8     80.2          79.2          -1.25%
BenchmarkParseIPPort/v6_v4-8           86.9          81.8          -5.87%
BenchmarkParseIPPort/v6_zone-8         176           169           -3.98%
```

Following recent @danderson twitter storm here is my take that is a bit faster. 
Not a big deal, just for your amusement. Probably could be improved by reading multiple bytes at once or vectorizing (not sure if possible) but at the cost if readability which definitely makes no sense in a grand scheme of things.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -163,58 +163,37 @@ func (err parseIPError) Error() string {
 }
 
 // parseIPv4 parses s as an IPv4 address (in form "192.168.0.1").
-func parseIPv4(in string) (ip IP, err error) {
-	s := in
-	var ip4 [4]byte
-
-	for i := 0; i < 4; i++ {
-		var (
-			j   int
-			acc uint16
-		)
-		// Parse one byte of digits. Bail if we overflow, stop at
-		// first non-digit.
-		//
-		// As of Go 1.15, don't try to factor this digit reading into
-		// a helper function. Its complexity is slightly too high for
-		// inlining, which ends up costing +50% in parse time.
-		for j = 0; j < len(s); j++ {
-			if s[j] < '0' || s[j] > '9' {
-				break
+func parseIPv4(s string) (ip IP, err error) {
+	var fields [3]uint8
+	var val, pos int
+	for i := 0; i < len(s); i++ {
+		if s[i] >= '0' && s[i] <= '9' {
+			val = val*10 + int(s[i]) - '0'
+			if val > 255 {
+				return IP{}, parseIPError{in: s, msg: "IPv4 field has value >255"}
 			}
-			acc = (acc * 10) + uint16(s[j]-'0')
-			if acc > 255 {
-				return IP{}, parseIPError{in: in, msg: "IPv4 field has value >255", at: s}
+		} else if s[i] == '.' {
+			// .1.2.3
+			// 1.2.3.
+			// 1..2.3
+			if i == 0 || i == len(s)-1 || s[i-1] == '.' {
+				return IP{}, parseIPError{in: s, msg: "IPv4 field must have at least one digit", at: s[i:]}
 			}
-		}
-		// There must be at least 1 digit per quad.
-		if j == 0 {
-			return IP{}, parseIPError{in: in, msg: "each dot-separated field must have at least one digit", at: s}
-		}
-
-		ip4[i] = uint8(acc)
-
-		// Non-final byte must be followed by a dot
-		if i < 3 {
-			if len(s) == j {
-				return IP{}, parseIPError{in: in, msg: "address too short"}
-			} else if s[j] != '.' {
-				return IP{}, parseIPError{in: in, msg: "unexpected character", at: s[j:]}
+			// 1.2.3.4.5
+			if pos == 3 {
+				return IP{}, parseIPError{in: s, msg: "IPv4 address too long"}
 			}
-			j++
+			fields[pos] = uint8(val)
+			pos++
+			val = 0
+		} else {
+			return IP{}, parseIPError{in: s, msg: "unexpected character", at: s[i:]}
 		}
-
-		// Advance to the next set of digits.
-		s = s[j:]
 	}
-	if len(s) != 0 {
-		if s[0] == '%' {
-			return IP{}, parseIPError{in: in, msg: "zone specifier not allowed with IPv4 addresses", at: s}
-		}
-		return IP{}, parseIPError{in: in, msg: "trailing garbage after address", at: s}
+	if pos < 3 {
+		return IP{}, parseIPError{in: s, msg: "IPv4 address too short"}
 	}
-
-	return IPv4(ip4[0], ip4[1], ip4[2], ip4[3]), nil
+	return IPv4(fields[0], fields[1], fields[2], uint8(val)), nil
 }
 
 // parseIPv6 parses s as an IPv6 address (in form "2001:db8::68").
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -222,6 +222,12 @@ func TestParseIP(t *testing.T) {
 		"1234",
 		// IPv4 with a zone specifier
 		"1.2.3.4%eth0",
+		// IPv4 field must have at least one digit
+		".1.2.3",
+		"1.2.3.",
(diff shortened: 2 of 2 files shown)


netaddr.go
```diff
@@ -163,58 +163,34 @@ func (err parseIPError) Error() string {
...
-				break
-			}
-			acc = (acc * 10) + uint16(s[j]-'0')
-			if acc > 255 {
-				return IP{}, parseIPError{in: in, msg: "IPv4 field has value >255", at: s}
+func parseIPv4(s string) (ip IP, err error) {
+	var ip4 [4]uint8
+	var a, b int
+	for i := 0; i < len(s); i++ {
+		if s[i] >= '0' && s[i] <= '9' {
```
AlexanderYastrebov, line 170: Playing with it I have found that subtracting `'0'` first and converting to uint eliminates one `cmp` from assembly and it only checks if value is below 9.

netaddr.go
```diff
@@ -163,58 +163,34 @@ func (err parseIPError) Error() string {
...
-			if len(s) == j {
-				return IP{}, parseIPError{in: in, msg: "address too short"}
-			} else if s[j] != '.' {
-				return IP{}, parseIPError{in: in, msg: "unexpected character", at: s[j:]}
+		} else if s[i] == '.' {
+			// .1.2.3
+			// 1.2.3.
+			// 1..2.3
+			// 1.2.3.4.5
+			if i == 0 || i == len(s)-1 || s[i-1] == '.' || b == 3 {
```
AlexanderYastrebov, line 180: Edge cases `i == 0 || i == len(s)-1` could be checked once before the loop but it does not bring any performance improvement.

netaddr.go
```diff
@@ -163,58 +163,34 @@ func (err parseIPError) Error() string {
...
-		// inlining, which ends up costing +50% in parse time.
-		for j = 0; j < len(s); j++ {
-			if s[j] < '0' || s[j] > '9' {
-				break
-			}
-			acc = (acc * 10) + uint16(s[j]-'0')
-			if acc > 255 {
-				return IP{}, parseIPError{in: in, msg: "IPv4 field has value >255", at: s}
+func parseIPv4(s string) (ip IP, err error) {
+	var ip4 [4]uint8
```
AlexanderYastrebov, line 167: I have also tried to use int64 and shift counter instead of byte array but that did not gave any improvement.

AlexanderYastrebov, line 167: I have also tried to use int64 and shift counter instead of byte array but that did not gave any improvement.

netaddr.go
danderson, line 180: The error message is wrong for b==3, it should say something about having too many fields.

danderson (approved):
Looks pretty good to me. One minor comment about error paths.

You also need to add a Signed-Off-By line if you accept the terms of the DCO. The text of the DCO is at https://developercertificate.org/ . If you're okay with that, you can `git commit --amend --signoff` to add the Signed-Off-By line to the commit.

Thanks!

netaddr.go
```diff
@@ -163,58 +163,34 @@ func (err parseIPError) Error() string {
...
+	var a, b int
```
bradfitz, line 168: maybe `ndot` for `b` at least?

netaddr.go
bradfitz, line 167: this could be `[3]uint8`, no?

netaddr.go
AlexanderYastrebov: Fixed

netaddr.go
AlexanderYastrebov: Fixed

netaddr.go
AlexanderYastrebov: Fixed

netaddr.go
```diff
@@ -163,58 +163,37 @@ func (err parseIPError) Error() string {
...
-		// inlining, which ends up costing +50% in parse time.
-		for j = 0; j < len(s); j++ {
-			if s[j] < '0' || s[j] > '9' {
-				break
+func parseIPv4(s string) (ip IP, err error) {
+	var fields [3]uint8
+	var val, pos int
+	for i := 0; i < len(s); i++ {
+		if s[i] >= '0' && s[i] <= '9' {
+			val = val*10 + int(s[i]) - '0'
```
AlexanderYastrebov, line 171: When brackets matter:
```go
val = val*10 + int(s[i]) - '0'
```
  
```asm
  55010a:       48 8d 34 b6             lea    rsi,[rsi+rsi*4] ; rsi = rsi*4 + rsi -> val*5
  55010e:       48 8d 34 77             lea    rsi,[rdi+rsi*2] ; rsi*2 + rdi       -> val*10 + s[i]
  550112:       48 8d 76 d0             lea    rsi,[rsi-0x30]  ; rsi = rsi - 0x30  -> val*10 + s[i] - '0'
```

```go
val = val*10 + int(s[i] - '0')
```

```asm
  55010a:       48 8d 34 b6             lea    rsi,[rsi+rsi*4] ; rsi = rsi*4 + rsi      -> val*5
  55010e:       83 c7 d0                add    edi,0xffffffd0  ; edi = edi + 0xffffffd0 -> s[i] - '0'
  550111:       40 0f b6 ff             movzx  edi,dil         ;                        -> int(s[i] - '0')
  550115:       48 8d 34 77             lea    rsi,[rdi+rsi*2] ; rsi = rsi*2 + rdi      -> val*10 + int(s[i] - '0')
```

mdlayher: Mind blowing. But probably worth leaving a comment that what you've written specifically compiles to less instructions. Or maybe there's room for a compiler optimization here @josharian? Beats me!

AlexanderYastrebov: Just an amusement I would not put a comment that we trick compiler here.
I would rather rely on benchmark numbers if this is optimized further or a regression is observed.

josharian: Can't do this in the compiler, because they have different overflow/underflow semantics. Picking the right order of operations can definitely help.

netaddr.go
josharian: The compiler should do this for you if you assign s[i] to a variable first (so it can see that it can safely re-evaluate the expression multiple times). That may be a bit more readable.

josharian: Another way to do this is to do something like:

```go
for i, b := range []byte(s) {
```

josharian: And for a fun twist, I believe the compiler will generate the same code if you do:

```go
switch s[i] {
case '0', '1', '2', '3', '4', '5', '6', '7', '8', '9':
  // etc
case '.':
  // etc
default:
  // etc
}
```

AlexanderYastrebov:
This one https://github.com/inetaf/netaddr/pull/95#issuecomment-752800277 shows how brittle all such optimizations are

josharian:
> This one #95 (comment) shows how brittle all such optimizations are

Sort of. But also not--it has a clear explanation (https://github.com/inetaf/netaddr/issues/102, which I just noticed independently). And some of the people who contribute to this repo know the compiler internals fairly well, and thus have a decent sense of what's likely to be brittle and what isn't. And if there is a regression in the compiler, we have the means to report it, fix it, and lock it in with a test.

danderson (approved):
I think this version is a bit more readable due to the absence of a nested loop, and it's also slightly faster (uint128 changes notwithstanding). Based on that, merging. Thanks!

PR #93, opened 2020-12-30
Remove dependency on stdlib for ParseIPPort, make faster.

danderson:
Makes long strings parse a little faster, because unlike net.SplitHostPort, we
can entirely ignore the contents of the ip and port portions, and let our subsequent
parser deal with errors there.

```
ParseIPPort/v4-8             56.5ns ± 3%    57.4ns ± 3%     ~     (p=0.421 n=5+5)
ParseIPPort/v6-8              143ns ± 2%     117ns ± 4%  -18.44%  (p=0.008 n=5+5)
ParseIPPort/v6_ellipsis-8     112ns ± 5%     100ns ± 1%  -10.68%  (p=0.008 n=5+5)
ParseIPPort/v6_v4-8           143ns ± 5%     139ns ± 1%     ~     (p=0.175 n=5+5)
ParseIPPort/v6_zone-8         272ns ± 1%     256ns ± 1%   -5.96%  (p=0.008 n=5+5)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -694,12 +694,41 @@ type IPPort struct {
 	Port uint16
 }
 
+// splitIPPort splits s into an IP address string and a port
+// string. It splits strings shaped like "foo:bar" or "[foo]:bar",
+// without further validating the substrings. v6 indicates whether the
+// ip string should parse as an IPv6 address or an IPv4 address, in
+// order for s to be a valid ip:port string.
+func splitIPPort(s string) (ip, port string, v6 bool, err error) {
+	i := strings.LastIndexByte(s, ':')
+	if i == -1 {
+		return "", "", false, errors.New("not an ip:port")
+	}
+
+	ip, port = s[:i], s[i+1:]
+	if len(ip) == 0 {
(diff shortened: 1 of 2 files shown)


danderson:
Note this adds a test case that diverges from the behavior of net: in netaddr.ParseIPPort, enclosing IPv4 addresses in brackets is not allowed, so `[1.2.3.4]:1234` will result in a parse error. Prior to this change we allowed this syntax because net.SplitHostPort doesn't care. That said, I _think_ brackets-with-IPv4 is supposed to not parse.

danderson:
Source: https://tools.ietf.org/html/rfc3986#section-3.2.2 . URI syntax says that square brackets are only valid for IPv6 or future protocols, and IPv4 must be a naked dotted-decimal. So, I think this new behavior (and deviation from stdlib) is correct?

bradfitz (approved):

josharian:
> So, I think this new behavior (and deviation from stdlib) is correct?

If we're deviating from package net and it's a bug fix, we should probably file an upstream bug so they can fix too.

PR #94, opened 2020-12-30
Replace PutUDPAddr with UDPAddrReuse.

danderson:
Exposing a sync.Pool across API boundaries is error-prone, as illustrated
by the dire warnings in PutUDPAddr's docstring.

Instead, split UDPAddr into two methods: one allocates and returns a UDPAddr,
the other reuses a passed-in UDPAddr. This allows the caller to do memory
management of the UDPAddrs as befits their application, makes netaddr failures
less mysterious (worst case, a nil dereference panic), and still lets high
performance applications get allocation-free behavior.

Why two methods rather than a single UDPAddr(reuse *net.UDPAddr) that allocates if reuse==nil? Because
combining the two functions and allocating on nil pushes the function
complexity *just* high enough to outline UDPAddr, resulting in a function
call performance hit. By having two accessors tailored to different uses,
both get inlined.

@bradfitz for your consideration. Will require some adjustment in the tailscale repo to update, but I feel more comfortable having the sync.Pool not cross API boundaries the way it does with PutUDPAddr. Thoughts?


netaddr.go
```diff
@@ -781,28 +780,28 @@ func FromStdAddr(stdIP net.IP, port int, zone string) (_ IPPort, ok bool) {
...
 }
 
-// PutUDPAddr adds ua to an internal pool for later reuse by IPPort.UDPAddr.
-// Use of PutUDPAddr is optional; improper use can cause mysterious errors.
-// You must only call PutUDPAddr if there are no remaining references to ua.
-func PutUDPAddr(ua *net.UDPAddr) { udpAddrPool.Put(ua) }
+// UDPAddrReuse is like UDPAddr, but reuses the provided UDPAddr,
+// which must be non-nil. If reuse.IP has a capacity of 16,
+// UDPAddrReuse is allocation-free.
+func (p IPPort) UDPAddrReuse(reuse *net.UDPAddr) *net.UDPAddr {
```
bradfitz, line 800: do we even need to return it? the caller has it. But then it needs a new name which probably doesn't sort next to UDPAddr. Hm, `UDPAddrPopulate`?

In any case, I much prefer having a separate method for the weird memory-reuse case rather than unifying them into one and forcing normal people to pass mystery `nil` values.

danderson: Returning it is nice for compactness when sending, the caller can `sock.WriteToUDP(bs, ipp.UDPAddr(reuse))`. I'm also fine with not returning if you find that too weird, but I think it's nice to make the method pass-through.

UDPAddrPopulate or UDPAddrFill sgtm if we don't return.

bradfitz: Yeah, it'll probably be more convenient. I'd just thought that the caller was going to need to do some sync.Pool work anyway, but it'd probably be like:

```go
    ua := pool.Get().(*net.UDPAddr)
    defer pool.Put(ua)
    sock.WriteToUDP(s, ipp.UDPAddrAt(ua))
```

And I'm thinking I like the name `UDPAddrAt`, like https://golang.org/pkg/reflect/#NewAt (like C++ "[placement new](https://en.cppreference.com/w/cpp/language/new#Placement_new)")

danderson: UDPAddrAt sgtm, renamed.

PR #96, opened 2020-12-30
Danderson/udpaddr noinline

danderson:
Unifies UDPAddr and UDPAddrReuse, which runs over the inlining limit by a tiny fraction. @josharian as requested. If you're able to get the function cost down from 84/80, we could have a single API method for UDPAddrs that can be convenient _and_ efficient.


netaddr.go
```diff
@@ -781,28 +780,20 @@ func FromStdAddr(stdIP net.IP, port int, zone string) (_ IPPort, ok bool) {
...
-	ua := udpAddrPool.Get().(*net.UDPAddr)
-	ua.Port = int(p.Port)
-	ua.IP, ua.Zone = p.IP.ipZone(ua.IP)
-	return ua
-}
-
-// PutUDPAddr adds ua to an internal pool for later reuse by IPPort.UDPAddr.
-// Use of PutUDPAddr is optional; improper use can cause mysterious errors.
-// You must only call PutUDPAddr if there are no remaining references to ua.
-func PutUDPAddr(ua *net.UDPAddr) { udpAddrPool.Put(ua) }
```
bradfitz, line 805: not a huge fan of breaking API on people. But I guess it might just be us.

netaddr.go
```diff
@@ -781,28 +780,20 @@ func FromStdAddr(stdIP net.IP, port int, zone string) (_ IPPort, ok bool) {
...
+// UDPAddr necessarily does two allocations. If you have an existing
+// UDPAddr already allocated, see UDPAddrReuse.
```
bradfitz, line 788: there is no UDPAddrReuse.

josharian:
On my phone, but what if you change IP.ipZone to:

```go
func (ip IP) ipZone(reuse net.IP) (stdIP net.IP, zone string) {
	base := reuse[:0]
	switch {
	case ip.z == z0:
	case ip.Is4():
		a4 := ip.As4()
		stdIP = append(base, a4[:]...)
	default:
		a16 := ip.As16()
		stdIP = append(base, a16[:]...)
      zone = ip.Zone()
	}
   return
}
```

netaddr.go
```diff
@@ -781,28 +780,20 @@ func FromStdAddr(stdIP net.IP, port int, zone string) (_ IPPort, ok bool) {
...
+func (p IPPort) UDPAddr(reuse *net.UDPAddr) *net.UDPAddr {
```
bradfitz, line 789: This feels like Hash.Sum (https://golang.org/pkg/hash/#Hash) which is always a bit meh. I think most people just pass nil (as they will here) and then the nils are mysterious when reading.

netaddr.go
bradfitz: Thanks, GitHub. I left that comment on the second commit where it didn't show up, but in the unified view my comment looks stupid.

josharian:
EBEDTIME

Will look again tomorrow.

netaddr.go
danderson: Sorry @bradfitz, I made this PR just gecause @josharian offered to help figure out how to get UDPAddr(reuse *net.UDPAddr) below the inlining threshold, and I was curious to see the technique at work. But if you prefer separate methods, consider this PR aborted and closed.

netaddr.go
danderson: Pretty sure it's just us. Google search seems to agree, only hits are tailscale and github forks of tailscale.

danderson:
@josharian No change to function complexity with the rewrite you suggested. I'm going to close this PR to keep our pending changes dash clean, and it looks like we're going to go with separate functions (which puts them both below the inlining threshold).

PR #97, opened 2020-12-30
Implement IP.String without relying on the stdlib.

danderson:
Makes String faster, and only makes 1 heap allocation for most IP shapes.
(v6+zone where the IP is the longest it can be can still cause extra allocs
to append the zone string)

IP.String is faster than net.IP.String for all cases, except IPv4-mapped IPv6,
because the stdlib prints it as an IPv4 and we print it as an IPv6, and so
need to do more work to format it.

```
name                    old time/op    new time/op    delta
IPString/v4-8              267ns ± 1%      53ns ± 3%  -80.03%  (p=0.000 n=9+10)
IPString/v6-8              268ns ± 1%     130ns ± 3%  -51.62%  (p=0.000 n=9+10)
IPString/v6_ellipsis-8     198ns ± 3%     117ns ± 3%  -40.74%  (p=0.000 n=10+10)
IPString/v6_v4-8           306ns ± 3%     121ns ± 2%  -60.32%  (p=0.000 n=9+10)
IPString/v6_zone-8         302ns ± 3%     135ns ± 1%  -55.24%  (p=0.000 n=10+10)

name                    old alloc/op   new alloc/op   delta
IPString/v4-8              16.0B ± 0%     16.0B ± 0%     ~     (all equal)
IPString/v6-8              48.0B ± 0%     48.0B ± 0%     ~     (all equal)
IPString/v6_ellipsis-8     32.0B ± 0%     32.0B ± 0%     ~     (all equal)
IPString/v6_v4-8           20.0B ± 0%     16.0B ± 0%  -20.00%  (p=0.000 n=10+10)
IPString/v6_zone-8         64.0B ± 0%     32.0B ± 0%  -50.00%  (p=0.000 n=10+10)

name                    old allocs/op  new allocs/op  delta
IPString/v4-8               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
IPString/v6-8               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
IPString/v6_ellipsis-8      1.00 ± 0%      1.00 ± 0%     ~     (all equal)
IPString/v6_v4-8            3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=10+10)
IPString/v6_zone-8          2.00 ± 0%      1.00 ± 0%  -50.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -410,6 +410,10 @@ func (ip IP) v6(i uint8) uint8 {
 	return uint8(ip.addr[(i/8)%2] >> ((7 - i%8) * 8))
 }
 
+func (ip IP) v6u16(i uint8) uint16 {
+	return uint16(ip.addr[(i/4)%2] >> ((3-i%4) * 16))
+}
+
 // IsZero reports whether ip is the zero value of the IP type.
 // The zero value is not a valid IP address of any type.
 //
@@ -557,7 +561,7 @@ func (ip IP) IsLinkLocalUnicast() bool {
 		return ip.v4(0) == 169 && ip.v4(1) == 254
 	}
 	if ip.Is6() {
-		return ip.v6(0) == 0xfe && ip.v6(1) == 0x80
+		return ip.v6u16(0) == 0xfe80
 	}
 	return false // zero value
 }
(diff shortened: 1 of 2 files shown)


danderson:
For comparison, net.IP.String on the same inputs and machine:

```
name                       time/op
IPStringStd/v4-8           65.8ns
IPStringStd/v6-8            251ns
IPStringStd/v6_ellipsis-8   184ns
IPStringStd/v6_v4-8        71.1ns
```

Number of allocs and alloc size is the same as netaddr.IP.String. v6_zone doesn't parse in stdlib, hence why it's missing from the list.

mdlayher (approved):
Nice work.

netaddr.go
```diff
@@ -406,6 +406,14 @@ func (ip IP) v6(i uint8) uint8 {
 	}
 }
 
+func (ip IP) v6_16(i uint8) uint16 {
```
bradfitz, line 409: if you change `_` to `u` I think it'll still okay and not have an underscore. And be a bit more informative?

netaddr.go
```diff
@@ -643,22 +651,108 @@ func (ip IP) As4() [4]byte {
...
+	if x >= 0x100 {
+		b = append(b, digits[x>>8&0xf])
+	}
+	if x >= 0x10 {
+		b = append(b, digits[x>>4&0xf])
+	}
+	return append(b, digits[x&0xf])
+}
+
+func (ip IP) string4() string {
```
bradfitz, line 699: sigh. I liked the simplicity of the fmt.Sprintf one-liner. alas.

netaddr.go
```diff
@@ -643,22 +651,108 @@ func (ip IP) As4() [4]byte {
...
+// to compact a single zero field.
+func (ip IP) string6() string {
+
+	zeroStart, zeroEnd := uint8(255), uint8(255)
+	for i := uint8(0); i < 8; i++ {
+		j := i
+		for j < 8 && ip.v6_16(j) == 0 {
+			j++
 		}
-		return fmt.Sprintf("::ffff:%x:%x%s%s", uint16(ip.v4(0))<<8|uint16(ip.v4(1)), uint16(ip.v4(2))<<8|uint16(ip.v4(3)), percent, zone)
```
bradfitz, line 658: is this a behavior change?

danderson: No behavior change, the string code no longer treats mapped IPs as special, so they get printed through the generic IPv6 printing code - which formats the same as this line. We had this special case previously to short-circuit the stdlib code, which formats these IPs as `::ffff:1.2.3.4`, but our code never does.

bradfitz: ah, I read that fmt.Sprintf too quickly... I was assuming _our_ special case previously was making the `::ffff:1.2.3.4` form. It was the opposite... we were avoiding it. Gotcha.

PR #99, opened 2020-12-30
Use uint128 for mask6 also.

bradfitz:


netaddr.go
```diff
@@ -1190,6 +1187,11 @@ func (r IPRange) Overlaps(o IPRange) bool {
 // Index 0 contains the high bits, index 1 contains the low.
 type uint128 [2]uint64
 
+// and returns the bitwise AND or u and m (u&m).
```
mdlayher, line 1190: *for

mdlayher (approved):

netaddr.go
bradfitz: Good catch. I think I meant "of" actually.

Issue #84, opened 2020-12-29
should IPRanges with different zones be valid?


josharian:
Should this be a valid range?

From: `::%a`

To: `::%b`

Seems pretty weird, but the current code allows it.

josharian:
Actually, the current code is even weirder, because zones play into the Less function. So From=::1%b, To=::2%a is valid, but From=::1%b, To=::1%a is not, because b > a.

josharian:
IPPrefix.Contains also ignores zones. Maybe that's right? Seems like maybe it'd be worth taking a pass over the whole package documenting what happens with zones, and making sure the implementation matches. I'm not sure I have enough networking experience to make those decisions well.

bradfitz:
Yeah, we definitely need a policy decision & docs on this.

How about:

* an `IPRange` is not `Valid` if the `From` & `To` zones don't match
* an `IPPrefix` does not contain an IP with a different zone
* an `IPRange` does not contain an IP with a different zone

/cc @danderson @mdlayher

mdlayher:
+1 from me, I've been thinking on this and I think the presence of differing zones should be enough to ensure that prefixes/ranges/IPs don't overlap.

mdlayher:
The canonical example would be link-local addresses in IPv6 which might be the same on different logical interfaces but do not overlap due to the zones:

```
[matt@routnerr-2:~]$ ip -6 a s | grep "fe80::" | sort | uniq -c
      1     inet6 fe80::20d:b9ff:fe53:eacc/64 scope link 
      6     inet6 fe80::20d:b9ff:fe53:eacd/64 scope link 
      1     inet6 fe80::/64 scope link 
```

danderson:
This got fixed a while back.

PR #100, opened 2020-12-30
make IPRange.Valid report false when zones differ

bradfitz:
Also probably faster implementation (by accident).

Updates #84 (more remains)

Signed-off-by: Brad Fitzpatrick <brad@danga.com>


mdlayher (approved):
LGTM.

PR #105, opened 2020-12-31
Use IPPrefix.Valid in other IPPrefix methods.

danderson:
Adjust some tests to check behavior on invalid prefixes.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -1034,14 +1034,13 @@ func (p IPPrefix) Range() IPRange {
 // The returned value is always non-nil.
 // Any zone identifier is dropped in the conversion.
 func (p IPPrefix) IPNet() *net.IPNet {
-	bits := 128
-	if p.IP.Is4() {
-		bits = 32
+	if !p.Valid() {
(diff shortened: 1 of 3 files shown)


danderson:
Caused a couple minor changes to inlining, but I'm treating those similar to depaware: updated the test to make us think about whether a particular outlining is bad or not, but not a showstopper.

netaddr.go
```diff
@@ -1034,10 +1034,14 @@ func MustParseIPPrefix(s string) IPPrefix {
 // Masked returns p in its canonical form, with bits of p.IP not in p.Bits masked off.
 // If p is zero or otherwise invalid, Masked returns the zero value.
 func (p IPPrefix) Masked() IPPrefix {
-	if m, err := p.IP.Prefix(p.Bits); err == nil {
-		return m
+	if !p.Valid() {
```
bradfitz, line 1037: why is this necessary? won't IP.Prefix return an error if it's invalid?

danderson: Doh, so it does. Reverted.

Issue #64, opened 2020-12-24
proposal: get rid of Unmap and *Raw


danderson:
In a twitter thread (end of the thread: https://twitter.com/dave_universetf/status/1341977567658536960), the question came up of why handle IPv4-mapped IPv6 addresses specially. After much consideration, I can't identify any good use cases for unmapping these addresses automatically in the general case. I think we should remove Unmap, and make the `*Raw` constructors the default and only constructors.

In more detail: in theory, IPv4-mapped IPv6 addresses are a way to cram an IPv4 address into the IPv6 address space. Notionally, the benefit of this is that you can represent both internets with a single address type. However, in practice we still distinguish the two internets in our code, and require our callers to care (via `Is4` et al.).

Per https://tools.ietf.org/html/rfc4038#section-4.2, the original purpose was to allow APIs and applications to wield v6 addresses only, and generate IPv4 packets further down in the network stack. In practice, this really hasn't happened, and the network stacks have remained dual-stack all the way up to the application layer.

Meanwhile, automatically unmapping introduces confusions: you might parse an IPv6 address, and find that `Is4()==true`. In the extreme, this could cause security issues if you're, say, writing a firewall. Alternatively, assuming that if you gave us an IPv6 address, you meant for it to be IPv6, offers a very consistent behavior.

Given that I'm failing to find a use for IPv4-mapped IPv6 at all, I think we should stop featuring it prominently in our API, and just remove it. If we were to support v4-in-v6 encoding, it would be much more useful to support the transition mechanisms that are actually used (but I also think we shouldn't put that in netaddr).

/cc @bradfitz @mdlayher would y'all miss IPv4-mapped IPv6 if it went away?

If we do this, one question is what the FromStd constructors should do, since they treat the v4-mapped 16 byte and the 4 byte representations of IPv4 interchangeably. I think for that case, given how people wield the stdlib, those constructors should always unmap (or maybe we keep offering a Raw constructor for the stdlib type only?). That way, IPs that would pass To4() in the stdlib become pure IPv4 in netaddr, and the rest become IPv6.

Past discussions of unmapping: https://github.com/inetaf/netaddr/issues/34

mdlayher:
> would y'all miss IPv4-mapped IPv6 if it went away?

Not one bit! I've always found the stdlib behaviors annoying.

> That way, IPs that would pass To4() in the stdlib become pure IPv4 in netaddr, and the rest become IPv6.

This seems like the correct choice to me.

[...]

PR #108, opened 2020-12-31
De-emphasize IPv4-mapped IPv6 handling, relegate it to stdlib compat.

danderson:
Proposal implementation for #64, to de-emphasize auto-unmapping. Keep it for stdlib because stdlib makes ambiguous types, but otherwise constructors don't do unobvious things.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -77,7 +77,7 @@ var (
 
 // IPv6LinkLocalAllNodes returns the IPv6 link-local all nodes multicast
 // address ff02::1.
-func IPv6LinkLocalAllNodes() IP { return IPv6Raw([16]byte{0: 0xff, 1: 0x02, 15: 0x01}) }
+func IPv6LinkLocalAllNodes() IP { return IPFrom16([16]byte{0: 0xff, 1: 0x02, 15: 0x01}) }
 
 // IPv6Unspecified returns the IPv6 unspecified address ::.
 func IPv6Unspecified() IP { return IP{z: z6noz} }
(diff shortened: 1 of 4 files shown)


mdlayher (approved):
LGTM

bradfitz:
I have feedback when at computer.

netaddr.go
```diff
@@ -90,10 +90,8 @@ func IPv4(a, b, c, d uint8) IP {
 	}
 }
 
-// IPv6Raw returns the IPv6 address given by the bytes in addr,
-// without an implicit Unmap call to unmap any v6-mapped IPv4
-// address.
-func IPv6Raw(addr [16]byte) IP {
+// IPFrom16 returns the IPv6 address given by the bytes in addr.
+func IPFrom16(addr [16]byte) IP {
```
bradfitz, line 94: This violates the "new behavior, new name" adage. It's better to break existing callers than silently break them with a change in behavior.

If this will only return IPv6 now, maybe a better name is just `IPv6` to match https://godoc.org/inet.af/netaddr#IPv4.

netaddr.go
```diff
@@ -1289,6 +1280,10 @@ func (u uint128) or(m uint128) uint128 {
 	return uint128{u.hi | m.hi, u.lo | m.lo}
 }
 
+func (u uint128) not() uint128 {
```
bradfitz, line 1283: the other three above have docs, even if internal

netaddr_test.go
```diff
@@ -337,12 +337,12 @@ func TestFromStdIP(t *testing.T) {
...
 			fn:   FromStdIP,
 			std:  net.ParseIP("::1"),
-			want: IPv6Raw([...]byte{15: 1}),
+			want: IPFrom16([...]byte{15: 1}),
 		},
 		{
 			name: "4in6-unmap",
 			fn:   FromStdIP,
-			std:  net.ParseIP("1.2.3.4"),
+			std:  net.ParseIP("1.2.3.4").To16(),
```
bradfitz, line 345: this change shouldn't be doing anything, right? I'd prefer you keep the old line alone and add a second test, even if it's currently redundant. It'd be interesting if it one of the cases failed in a future Go release.

PR #114, opened 2021-01-01
fun with bit twiddling

josharian:
A bit of fun for me to wrap up 2020. I'm not wedded to these, so if you think it's too much, we can revert.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -1070,10 +1070,11 @@ func (p IPPrefix) Contains(ip IP) bool {
 		// If all the bits we care about are equal, the result will be zero.
 		return uint32((ip.addr.lo^p.IP.addr.lo)>>((32-p.Bits)&63)) == 0
 	} else {
+		// xor the IP addresses together.
+		// Mask away the bits we don't care about.
+		// If all the bits we care about are equal, the result will be zero.
 		m := mask6(p.Bits)
-		// TODO: benchmark whether the short circuit below is faster or slower
-		// than the higher level alternative: 'ip.addr.and(m) == p.IP.addr.and(m)'.
-		return ip.addr.hi&m.hi == p.IP.addr.hi&m.hi && ip.addr.lo&m.lo == p.IP.addr.lo&m.lo
+		return ip.addr.xor(p.IP.addr).and(m).isZero()
 	}
 }
 
--- a/netaddr.go
+++ b/netaddr.go
@@ -1037,12 +1037,6 @@ func (p IPPrefix) IPNet() *net.IPNet {
 	}
 }
 
-// mask4 returns a bit mask that selects the topmost n bits of a
-// uint32.
-func mask4(n uint8) uint32 {
-	return ^uint32(0) << (32 - n)
-}
-
 // mask6 returns a bit mask that selects the topmost n bits of a
 // 128-bit number.
 func mask6(n uint8) uint128 {
@@ -1066,8 +1060,15 @@ func (p IPPrefix) Contains(ip IP) bool {
 		return false
 	}
 	if ip.Is4() {
-		m := mask4(p.Bits)
-		return uint32(ip.addr.lo)&m == uint32(p.IP.addr.lo)&m
+		// xor the IP addresses together; mismatched bits are now ones.
+		// Shift away the number of bits we don't care about.
+		// Shifts in Go are more efficient if the compiler can prove
+		// that the shift amount is smaller than the width of the shifted type (64 here).
+		// We know that p.Bits is in the range 0..32 because p is Valid;
+		// the compiler doesn't know that, so mask with 63 to help it.
+		// Now truncate to 32 bits, because this is IPv4.
+		// If all the bits we care about are equal, the result will be zero.
+		return uint32((ip.addr.lo^p.IP.addr.lo)>>((32-p.Bits)&63)) == 0
 	} else {
 		m := mask6(p.Bits)
 		// TODO: benchmark whether the short circuit below is faster or slower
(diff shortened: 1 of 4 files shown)


netaddr.go
```diff
@@ -1066,8 +1060,13 @@ func (p IPPrefix) Contains(ip IP) bool {
...
 	}
 	if ip.Is4() {
-		m := mask4(p.Bits)
-		return uint32(ip.addr.lo)&m == uint32(p.IP.addr.lo)&m
+		// xor the IP addresses together; mismatched bits are now ones.
+		// Shift away the number of bits we don't care about.
+		// We know that p.Bits is in the range 0..32 because p is Valid;
+		// the compiler doesn't know that, so mask with 63 to help it.
+		// Now truncate to 32 bits, because this is IPv4.
+		// If all the bits we care about are equal, the result wil be zero.
```
bradfitz, line 1068: typo: will.

netaddr.go
```diff
@@ -1068,10 +1068,11 @@ func (p IPPrefix) Contains(ip IP) bool {
 		// If all the bits we care about are equal, the result wil be zero.
 		return uint32((ip.addr.lo^p.IP.addr.lo)>>((32-p.Bits)&63)) == 0
 	} else {
+		// xor the IP addresses together.
+		// Mask away the bits we don't care about.
+		// If all the bits we care about are equal, the result wil be zero.
```
bradfitz, line 1073: same typo

uint128.go
```diff
@@ -47,14 +47,7 @@ func (u uint128) addOne() uint128 {
...
 func u64CommonPrefixLen(a, b uint64) uint8 {
-	for i := uint8(0); i < 64; i++ {
-		if a == b {
-			return 64 - i
-		}
-		a >>= 1
-		b >>= 1
-	}
-	return 0
+	return uint8(bits.LeadingZeros64(a ^ b))
```
bradfitz, line 50: gahh! I knew this trick and forgot it when I wrote this code. Thanks!

bradfitz (approved):
LGTM

uint128.go
danderson: Last night I re-derived this from first principles, then promptly forgot to implement it. Thanks Josh!

netaddr.go
```diff
@@ -1066,13 +1060,19 @@ func (p IPPrefix) Contains(ip IP) bool {
...
 	if ip.Is4() {
-		m := mask4(p.Bits)
-		return uint32(ip.addr.lo)&m == uint32(p.IP.addr.lo)&m
+		// xor the IP addresses together; mismatched bits are now ones.
+		// Shift away the number of bits we don't care about.
+		// We know that p.Bits is in the range 0..32 because p is Valid;
+		// the compiler doesn't know that, so mask with 63 to help it.
+		// Now truncate to 32 bits, because this is IPv4.
+		// If all the bits we care about are equal, the result wil be zero.
+		return uint32((ip.addr.lo^p.IP.addr.lo)>>((32-p.Bits)&63)) == 0
```
danderson, line 1069: I do want to call out that this feels significantly less readable than the prior code, for a 6% gain. However, this is also my kind of IP address manipulation code, so I'm all for it :P

netaddr.go
```diff
@@ -1066,13 +1060,19 @@ func (p IPPrefix) Contains(ip IP) bool {
...
 	} else {
+		// xor the IP addresses together.
+		// Mask away the bits we don't care about.
+		// If all the bits we care about are equal, the result wil be zero.
 		m := mask6(p.Bits)
-		// TODO: benchmark whether the short circuit below is faster or slower
-		// than the higher level alternative: 'ip.addr.and(m) == p.IP.addr.and(m)'.
-		return ip.addr.hi&m.hi == p.IP.addr.hi&m.hi && ip.addr.lo&m.lo == p.IP.addr.lo&m.lo
+		return ip.addr.xor(p.IP.addr).and(m).isZero()
```
danderson, line 1075: Vaguely surprised that you can't make it faster using the same tricks as v4. I would have expected it to scale fairly well to two parallel uint64 operations on OOO hardware. But, benchmarks don't lie I guess.

danderson (approved):

netaddr.go
josharian: The trouble is implementing uint128.shr efficiently. You need a branch, and an extra shift and or.

In the last hour I’ve started contemplating a proposal to add new API to math/bits, something like:

```go
// ShiftInRight64 right shifts lo by n bits.
// The lower n bits of hi are shifted in to replace them.
// If n is greater than 64, any extra bits shifted in are all zero.
func ShiftInRight64(hi, lo uint64, n uint8) uint64
```

Plus obvious friends to fill out the family (different bit widths, ShiftInLeft).

This would correspond to amd64’s SHRD instruction, plus some fix-up code for large shifts.

I don’t know what n’s type should be. int, uint, and uint8 are all natural choices.

math/big could then use this to implement its shrVU_g and shlVU_g primitives, maybe efficiently enough to allow us to delete the amd64 assembly.

netaddr.go
```diff
@@ -1066,13 +1060,19 @@ func (p IPPrefix) Contains(ip IP) bool {
 		return false
 	}
 	if ip.Is4() {
-		m := mask4(p.Bits)
-		return uint32(ip.addr.lo)&m == uint32(p.IP.addr.lo)&m
+		// xor the IP addresses together; mismatched bits are now ones.
+		// Shift away the number of bits we don't care about.
+		// We know that p.Bits is in the range 0..32 because p is Valid;
+		// the compiler doesn't know that, so mask with 63 to help it.
```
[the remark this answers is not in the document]
josharian: Maybe I haven't been hex-indoctrinated enough, but 63 seems more clearly 64-1 to me than 0x3f. I added another sentence earlier to clarify.

PR #117, opened 2021-01-04
Change mask6 to a table, simplify some other code.

bradfitz:


mask6.go
```diff
@@ -0,0 +1,141 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package netaddr
+
+// mask6 are bitmasks with the topmost n bits of a
+// 128-bit number, where n is the array index.
+//
+// generated with https://play.golang.org/p/dGA3h0jrTry
+var mask6 = [...]uint128{
```
bradfitz, line 11: maybe this should be a const string instead? the compiler might do a better job.

josharian: We don't do constant-foldable lookups. I'm not sure it's worth the considerable reduction in readability. I'm open to being convinced otherwise by numbers, though.

mdlayher (approved):
LGTM, I like the simplifications.

josharian (approved):
Nice. @danderson might decide to make uint128s native-endian once this is in, though. :P

I expect this'll be good for benchmarks, but wouldn't hurt to do a run to make sure.

PR #124, opened 2021-01-27
Reimplement IPSet.Ranges with a (somewhat) easier to follow algorithm.

danderson:
```
    The main goal was to fix #125 by providing a less tricky implementation
    of Ranges as a comparison baseline... But this implementation turns out
    to also be a bit faster and more memory efficient.
    
    Fixes #125.
    
    name         old time/op    new time/op    delta
    IPSetFuzz-8    16.3µs ± 3%    14.1µs ± 1%  -12.99%  (p=0.008 n=5+5)
    
    name         old alloc/op   new alloc/op   delta
    IPSetFuzz-8    2.60kB ± 0%    1.95kB ± 0%  -25.24%  (p=0.008 n=5+5)
    
    name         old allocs/op  new allocs/op  delta
    IPSetFuzz-8      33.0 ± 0%      30.0 ± 0%   -9.09%  (p=0.008 n=5+5)
    
    Signed-off-by: David Anderson <dave@natulte.net>
```

PR #126, opened 2021-01-27
Implement IPSet.Invert and IPSet.Intersect.

### LINKED ISSUE #123: IPSet needs more set operations
While implementing a complicated thing, I would like IPSet.Intersection(a,b) (returns a new IPSet containing only the things in both a and b). And if we're implementing one set operation, the others are probably worth having also (union, symmetric difference).

/cc @bradfitz for thoughts.

danderson:
Fixes #123.

I set the merge-base to #124 so it shows only the new functionality.

### DIFF
--- a/ipset.go
+++ b/ipset.go
@@ -22,6 +22,21 @@ type IPSet struct {
 	out []IPRange
 }
 
+// toInOnly updates s to clear s.out, by merging any s.out into s.in.
+func (s *IPSet) toInOnly() {
+	if len(s.out) > 0 {
+		s.in = s.Ranges()
+		s.out = nil
+	}
+}
+
+// Clone returns a copy of s that shares no memory with s.
+func (s *IPSet) Clone() *IPSet {
+	return &IPSet{
+		in: s.Ranges(),
+	}
+}
+
 // Add adds ip to the set s.
 func (s *IPSet) Add(ip IP) { s.AddRange(IPRange{ip, ip}) }
 
@@ -35,10 +50,7 @@ func (s *IPSet) AddRange(r IPRange) {
 	}
 	// If there are any removals (s.out), then we need to compact the set
 	// first to get the order right.
-	if len(s.out) > 0 {
-		s.in = s.Ranges()
-		s.out = nil
-	}
+	s.toInOnly()
 	s.in = append(s.in, r)
 }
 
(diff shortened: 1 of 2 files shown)


ipset.go
```diff
@@ -99,6 +112,23 @@ func (s *IPSet) RemoveSet(b *IPSet) {
 	}
 }
 
+// Invert updates s to contain the opposite of is current contents.
+func (s *IPSet) Invert() {
+	s.toInOnly()
+	s.out = s.in
+	s.in = []IPRange{
+		IPPrefix{IP: IPv4(0,0,0,0), Bits: 0}.Range(),
```
mdlayher, line 120: Is this gofmt'd? I'd expect spaces between the IPv4 parameters.

mdlayher (approved):
LGTM!

ipset.go
danderson: Oops, fixed. My NixOS was missing goimports in the env.

ipset.go
```diff
@@ -22,6 +22,22 @@ type IPSet struct {
...
+	if len(s.out) > 0 {
+		s.in = s.Ranges()
+		s.out = nil
+	}
+}
+
+// Clone returns a copy of s that shares no memory with s.
+func (s *IPSet) Clone() *IPSet {
+	return &IPSet{
+		in:  append([]IPRange(nil), s.in...),
```
josharian, line 36: Could opportunistically do

```go
return &IPSet{in: s.Ranges()}
```

ipset.go
```diff
@@ -99,6 +112,23 @@ func (s *IPSet) RemoveSet(b *IPSet) {
 	}
 }
 
+// Invert updates s to contain the opposite of is current contents.
```
josharian, line 115: Consider calling this `Complement`, since that's the standard set theory name for it.

s/is/its/

ipset.go
```diff
@@ -99,6 +112,23 @@ func (s *IPSet) RemoveSet(b *IPSet) {
...
+func (s *IPSet) Invert() {
+	s.toInOnly()
+	s.out = s.in
+	s.in = []IPRange{
+		IPPrefix{IP: IPv4(0, 0, 0, 0), Bits: 0}.Range(),
+		IPPrefix{IP: IPv6Unspecified(), Bits: 0}.Range(),
```
josharian, line 121: Should we iterate through the `s.in`, discover all IPv6 zones contained therein, and add those here? Should we have IPv6 at all if `s.in` only contains IPv4?

Either way, we should document what universe we're taking the complement in.

ipset.go
danderson: Doh, I knew I was using the wrong word. Thanks!

ipset.go
danderson: Hmmm. IPSet doesn't really deal with zones at all right now (e.g. compaction will smush together IPs across zones), so I think I'm going to turn this into a bug to figure out what to do, and we can fix all of IPSet at once. Until then, leaving as-is.

Issue #125, opened 2021-01-27
IPSet.Ranges sometimes generates invalid IPRanges


danderson:
Found while implementing other functionality and writing tests. PR at #124 adds an assert that trips on the existing single_ips test.

```
--- FAIL: TestIPSet (0.00s)
    --- FAIL: TestIPSet/single_ips (0.00s)
        ipset.go:162: post-sort:
        ipset.go:142:  {  10.0.0.0        ✅
        ipset.go:144:   } 10.0.0.0        ✅
        ipset.go:142:  {  10.0.0.1        ✅
        ipset.go:144:   } 10.0.0.1        ✅
        ipset.go:142:  {  10.0.0.2        ✅
        ipset.go:144:   } 10.0.0.2        ✅
        ipset.go:142:  {  10.0.0.3        ✅
        ipset.go:144:   } 10.0.0.3        ✅
        ipset.go:142:  {  10.0.0.4        ❌
        ipset.go:142:  {  10.0.0.4        ✅
        ipset.go:144:   } 10.0.0.4        ❌
        ipset.go:144:   } 10.0.0.4        ✅
        ipset.go:164: merging...
        ipset.go:183: at[0] ({ip:{addr:{hi:0 lo:281470849515520} z:0xc000128060} want:true start:true}), add=1, remove=0
        ipset.go:183: at[1] ({ip:{addr:{hi:0 lo:281470849515520} z:0xc000128060} want:true start:false}), add=0, remove=0
        ipset.go:183: at[2] ({ip:{addr:{hi:0 lo:281470849515521} z:0xc000128060} want:true start:true}), add=1, remove=0
        ipset.go:183: at[3] ({ip:{addr:{hi:0 lo:281470849515521} z:0xc000128060} want:true start:false}), add=0, remove=0
        ipset.go:183: at[4] ({ip:{addr:{hi:0 lo:281470849515522} z:0xc000128060} want:true start:true}), add=1, remove=0
        ipset.go:183: at[5] ({ip:{addr:{hi:0 lo:281470849515522} z:0xc000128060} want:true start:false}), add=0, remove=0
        ipset.go:183: at[6] ({ip:{addr:{hi:0 lo:281470849515523} z:0xc000128060} want:true start:true}), add=1, remove=0
        ipset.go:183: at[7] ({ip:{addr:{hi:0 lo:281470849515523} z:0xc000128060} want:true start:false}), add=0, remove=0
[...]

Issue #110, opened 2020-12-31
document concurrency properties


bradfitz:
We should document what methods are safe to call concurrently.

In general things aren't safe (and that's the default to assume in Go when not documented otherwise), but we have plenty of stuff that is safe to call concurrently. Anything on a value type, for instance. And the func returned by IPSet.ContainsFunc.

But also IPPrefix.Contains and IPRange.Contains and IPPrefix.Overlaps and IPRange.Overlaps. **But not** IPSet.Overlaps if #109 goes in.

So maybe we should decide what rules we want before we decide to submit #109.

/cc @danderson @josharian @mdlayher

josharian:
I'd love to be able to say that everything in the package is concurrency-safe. It makes it really, really nice to use, particularly for low-level types.

If that bar is unreasonable to clear, I'd vote to split that package into concurrency-safe types (probably the simplest, lowest-level types  like OP, IPPrefix, IPPort) and non-concurrency-safe types, and make no guarantees whatsoever about the latter ones.

josharian:
I'd like to come to a decision here. The menu of options seems to be:

1. Declare that value types (IP, IPRange, IPPrefix) are concurrency safe. Reference types (IPSet) are not.
2. Declare that all types are concurrency safe. Add a mutex to IPSet.
3. Declare that no types are concurrency safe.

My 2c:

3 seems like the worst option.

Given that we can backwards-compatibly go from 1 to 2, 1 seems the lower-risk option of the two. Mutex are cheap, though, so I'm also OK with 2.

danderson:
I was assuming 1 as our default. Concurrency safety is fantastic for consumers, so we should strive for it.

I'm on the fence about 2, but as long as we don't have any callback-based functions, I think throwing a mutex on it and guaranteeing that all types are concurrency-safe LGTM.

Agree on 3, let's not do that.

PR #109, opened 2020-12-31
Implement IPSet.Overlaps.

danderson:
Closes #81.

Signed-off-by: David Anderson <dave@natulte.net>

PR #127, opened 2021-01-28
add IPSet.Equal

josharian:
This depends on the bug fixes in https://github.com/inetaf/netaddr/pull/124 for its tests to pass.

### DIFF
--- a/ipset_test.go
+++ b/ipset_test.go
@@ -701,3 +701,35 @@ func TestIPSetRangesStress(t *testing.T) {
 		}
 	}
 }
+
+func TestIPSetEqual(t *testing.T) {
+	a := new(IPSet)
+	b := new(IPSet)
+
+	assertEqual := func(want bool) {
+		t.Helper()
+		if got := a.Equal(b); got != want {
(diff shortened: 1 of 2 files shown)


ipset_test.go
```diff
@@ -624,3 +624,35 @@ func TestPointLess(t *testing.T) {
...
+	}
+
+	a.Add(MustParseIP("1.1.1.0"))
+	a.Add(MustParseIP("1.1.1.1"))
+	a.Add(MustParseIP("1.1.1.2"))
+	b.AddPrefix(MustParseIPPrefix("1.1.1.0/31"))
+	b.Add(MustParseIP("1.1.1.2"))
+	assertEqual(true)
+
+	a.RemoveSet(a)
```
josharian, line 646: This effectively zeros `a`. But the underlying representation contains non-empty `in` and `out`. Fixing this is in part the topic of #109 and #110 and a decision about when to minimize the representation.

danderson (approved):

danderson:
Tests are failing, but I suspect it's because of #125, which should go away if you rebase.

PR #130, opened 2021-01-29
Make IPSet concurrency-safe.

danderson:
Addresses the non-documentation bits of #110.

### DIFF
--- a/ipset.go
+++ b/ipset.go
@@ -4,7 +4,10 @@
 
 package netaddr
 
-import "sort"
+import (
+	"sort"
+	"sync"
+)
 
 // IPSet represents a set of IP addresses.
 //
@@ -15,19 +18,13 @@ import "sort"
 // nothing on an empty set. Ranges may be fully, partially, or not
 // overlapping.
 type IPSet struct {
+	mu sync.Mutex // protects all fields
+
 	// in are the ranges in the set.
 	in []IPRange
 
 	// out are the ranges to be removed from 'in'.
 	out []IPRange
-}
-
-// toInOnly updates s to clear s.out, by merging any s.out into s.in.
-func (s *IPSet) toInOnly() {
-	if len(s.out) > 0 {
-		s.in = s.Ranges()
-		s.out = nil
-	}
 }
 
 // Clone returns a copy of s that shares no memory with s.
@@ -45,12 +42,20 @@ func (s *IPSet) AddPrefix(p IPPrefix) { s.AddRange(p.Range()) }
 
 // AddRange adds r to s.
 func (s *IPSet) AddRange(r IPRange) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.addRangeLocked(r)
+}
+
+func (s *IPSet) addRangeLocked(r IPRange) {
 	if !r.Valid() {
 		return
 	}
 	// If there are any removals (s.out), then we need to compact the set
 	// first to get the order right.
-	s.toInOnly()
+	if len(s.out) > 0 {
+		s.rangesLocked()
+	}
 	s.in = append(s.in, r)
 }
 
@@ -59,7 +64,9 @@ func (s *IPSet) Remove(ip IP) { s.RemoveRange(IPRange{ip, ip}) }
 
 // RemoveFreePrefix removes and returns a Prefix of length bits from the IPSet.
 func (s *IPSet) RemoveFreePrefix(bitLen uint8) (p IPPrefix, ok bool) {
-	prefixes := s.Prefixes()
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	prefixes := s.prefixesLocked()
 	if len(prefixes) == 0 {
 		return IPPrefix{}, false
 	}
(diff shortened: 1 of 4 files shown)


ipset.go
```diff
@@ -15,25 +18,24 @@ import "sort"
 // nothing on an empty set. Ranges may be fully, partially, or not
 // overlapping.
 type IPSet struct {
+	mu sync.Mutex
```
josharian, line 21: // protects all fields

or the like

ipset.go
```diff
@@ -154,6 +182,16 @@ var debugf = discardf
 // Ranges returns the minimum and sorted set of IP
 // ranges that covers s.
 func (s *IPSet) Ranges() []IPRange {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.rangesLocked()
+}
+
+func (s *IPSet) rangesLocked() []IPRange {
```
josharian, line 190: I’d call this normalize, and then let callers use s.in after calling it. That makes it clearer that it mutates s.

ipset.go
```diff
@@ -15,25 +18,24 @@ import "sort"
...
 
-// toInOnly updates s to clear s.out, by merging any s.out into s.in.
-func (s *IPSet) toInOnly() {
-	if len(s.out) > 0 {
-		s.in = s.Ranges()
-		s.out = nil
-	}
+	// normalized indicates that 'in' is in sorted order, and 'out' is
+	// empty.
+	normalized bool
```
josharian, line 31: How big of a performance win does this provide? Keeping this extra bit of state around. Could we calculate it on the fly cheaply when we need it?

ipset.go
```diff
@@ -92,34 +105,49 @@ func (s *IPSet) RemovePrefix(p IPPrefix) { s.RemoveRange(p.Range()) }
...
 		s.out = append(s.out, r)
+		s.normalized = false
 	}
 }
 
 // AddSet adds all ranges in b to s.
 func (s *IPSet) AddSet(b *IPSet) {
-	for _, r := range b.Ranges() {
-		s.AddRange(r)
+	rr := b.Ranges()
```
josharian, line 122: Why not rangesLock and moved mutex higher? I’d expect any given method to be atomic.

ipset.go
```diff
@@ -92,34 +105,49 @@ func (s *IPSet) RemovePrefix(p IPPrefix) { s.RemoveRange(p.Range()) }
...
+	for _, r := range rr {
+		s.addRangeLocked(r)
 	}
 }
 
 // RemoveSet removes all ranges in b from s.
 func (s *IPSet) RemoveSet(b *IPSet) {
-	for _, r := range b.Ranges() {
-		s.RemoveRange(r)
+	rr := b.Ranges()
```
josharian, line 132: Same—rangesLocked?

josharian (commented):
First look from my phone. Will review again on a computer tomorrow.

Can we add a stress test that concurrently mutates an IPSet using all different methods? That way running with -race is likely to catch accidental regressions.

ipset.go
danderson: Using s.in is a bit weird, because unless you call normalize first, it's likely that you'll get the wrong answer. My goal was to encourage use of "what do you mean" accessors as much as possible. That, combined with the fact that Ranges has to do the computations required to normalize, made me combine in this way. WDYT?

ipset.go
danderson: Didn't measure, it's possible this is pointless. Could make ranges smarter to detect an already-sorted s.in and empty s.out, so it can fast-exit after a single pass through s.in. Probably better, I'll alter to do that.

ipset.go
danderson: Lock ordering. If you want AddSet to be completely atomic, we'd have to simultaneously lock s and b, and that feels really easy to misuse in the caller (e.g. two goroutines, one calling a.AddSet(b) and another calling b.AddSet(a) could deadlock).

Given that, I chose to go with the semantics that _some_ consistent snapshot of b will be used, but not make promises beyond that.

ipset.go
danderson: Same as above, lock ordering makes this tricky.

ipset.go
danderson: Removed normalized, and instead added a mild fast path to rangesLocked that avoids allocating if it can. With that, rangesLocked still costs a slice sort plus one pass through the array, but not an alloc+copy unless removals are required.

bradfitz:
I'm not a huge fan of this.

I think only the read-only methods post construction need to be safe for concurrent use.

What does it mean to concurrently `Add` a range (say) 1.1.1.1-1.1.1.255 and `Remove` 1.1.1.100-1.1.1.200? That used to be a data race (good!) but now it's ill-defined randomness (but without a data race, so gold star?).

mdlayher:
This is no longer necessary now that we have IPSetBuilder and IPSet is immutable, yeah?

mdlayher:
99% sure this is obsolete with the addition of the mutable builder type. Closing.

PR #135, opened 2021-02-03
netaddr: assert more functions can be inlined

mdlayher:

These all inline as of Go 1.15.7.

```
=== CONT  TestInlining                                                                           
    inlining_test.go:110: not in expected set, but also inlinable: "IPRange.entirelyBefore"      
    inlining_test.go:110: not in expected set, but also inlinable: "(*IPSet).Clone"              
    inlining_test.go:110: not in expected set, but also inlinable: "(*IPSet).toInOnly"           
    inlining_test.go:110: not in expected set, but also inlinable: "IP.lessOrEq"                 
--- PASS: TestInlining (0.24s)
```

### DIFF
--- a/inlining_test.go
+++ b/inlining_test.go
@@ -43,8 +43,10 @@ func TestInlining(t *testing.T) {
 	})
 	for _, want := range []string{
 		"(*IPSet).Add",
+		"(*IPSet).Clone",
 		"(*IPSet).Remove",
 		"(*IPSet).RemoveRange",
+		"(*IPSet).toInOnly",
 		"(*uint128).halves",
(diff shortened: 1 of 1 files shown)


josharian:
Is the fact that these inline important for performance?

mdlayher:
No, I guess I just assumed that we were adding all of the possible functions to the list to keep things speedy without unintentional regression when making future changes. No worries if not.

josharian:
My 2c at least is to only add functions that we care about. Otherwise it's just noise when refactoring, etc. But I don't feel strongly.

mdlayher:
I'll just close this for now.

PR #136, opened 2021-02-03
netaddr: add IP.StringExpanded to expand IPv6 address strings

### LINKED ISSUE #51: proposal: add additional methods to support IPv6 expansion
Hi There,

I saw this package and thought that it overlapped with some work that I have been doing recently. With IPv6 the addresses are usually shortened, have you thought about adding a feature where you can print out the expanded IPv6 format?

Example
```
ip, _ := netaddr.ParseIP("2404:6800:4006:809::200e")
ip.As16ExpandedShort() // 2404:6800:4006:0809:0:0:0:200e
ip.As16ExpandedLong() // 2.4.0.4.6.8.0.0.4.0.0.6.0.8.0.9.0.0.0.0.0.0.0.0.0.0.0.0.2.0.0.e
```

The use case for this is the automatic creation of DNS IPv6 PTR records however I am sure that there are other use cases. Is this something that you have thought about as an addition to this package or do you see it as out of scope?

If it is in scope I am happy to do the work and provide you with a PR, I just wanted to check if its aligned with where you want to take your package.

mdlayher:

Fixes #51. There's a bit of duplication between `IP.string6` and `IP.StringExpanded` but I was able to remove some logic regarding adding `::` compression so it seemed worth duplicating.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -756,6 +756,11 @@ func appendHex(b []byte, x uint16) []byte {
 	return append(b, digits[x&0xf])
 }
 
+// appendHexPad appends the fully padded hex string representation of x to b.
+func appendHexPad(b []byte, x uint16) []byte {
+	return append(b, digits[x>>12], digits[x>>8&0xf], digits[x>>4&0xf], digits[x&0xf])
+}
+
 func (ip IP) string4() string {
 	const max = len("255.255.255.255")
 	ret := make([]byte, 0, max)
@@ -824,6 +829,34 @@ func (ip IP) appendTo6(ret []byte) []byte {
 		ret = append(ret, ip.Zone()...)
 	}
 	return ret
+}
+
+// StringExpanded is like String but IPv6 addresses are expanded with leading
+// zeroes and no "::" compression. For example, "2001:db8::1" becomes
+// "2001:0db8:0000:0000:0000:0000:0000:0001".
+func (ip IP) StringExpanded() string {
+	switch ip.z {
+	case z0, z4:
+		return ip.String()
+	}
+
+	const size = len("ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")
+	ret := make([]byte, 0, size)
+	for i := uint8(0); i < 8; i++ {
+		if i > 0 {
+			ret = append(ret, ':')
+		}
+
+		ret = appendHexPad(ret, ip.v6u16(i))
+	}
+
+	if ip.z != z6noz {
+		// The addition of a zone will cause a second allocation, but when there
+		// is no zone the ret slice will be stack allocated.
+		ret = append(ret, '%')
+		ret = append(ret, ip.Zone()...)
+	}
+	return string(ret)
 }
 
 // MarshalText implements the encoding.TextMarshaler interface,
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -784,6 +784,41 @@ func TestLessCompare(t *testing.T) {
 	}
 }
 
+func TestIPStringExpanded(t *testing.T) {
+	tests := []struct {
+		ip IP
+		s  string
+	}{
+		{
+			ip: IP{},
+			s:  "invalid IP",
+		},
+		{
+			ip: mustIP("192.0.2.1"),
+			s:  "192.0.2.1",
+		},
+		{
+			ip: mustIP("2001:db8::1"),
+			s:  "2001:0db8:0000:0000:0000:0000:0000:0001",
+		},
+		{
+			ip: mustIP("2001:db8::1%eth0"),
+			s:  "2001:0db8:0000:0000:0000:0000:0000:0001%eth0",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.ip.String(), func(t *testing.T) {
+			want := tt.s
+			got := tt.ip.StringExpanded()
+
+			if got != want {
+				t.Fatalf("got %s, want %s", got, want)
+			}
+		})
+	}
+}
+
 func TestIPPrefixMasking(t *testing.T) {
 	type subtest struct {
 		ip   IP
@@ -1879,6 +1914,18 @@ func BenchmarkIPString(b *testing.B) {
 			b.ReportAllocs()
 			for i := 0; i < b.N; i++ {
 				sinkString = ip.String()
+			}
+		})
+	}
+}
+
+func BenchmarkIPStringExpanded(b *testing.B) {
+	for _, test := range parseBenchInputs {
+		ip := MustParseIP(test.ip)
+		b.Run(test.name, func(b *testing.B) {
+			b.ReportAllocs()
+			for i := 0; i < b.N; i++ {
+				sinkString = ip.StringExpanded()
 			}
 		})
 	}


mdlayher:
Any thoughts?

netaddr.go
```diff
@@ -720,7 +732,38 @@ func (ip IP) string6() string {
...
+// See the documentation of IP.String for more details.
+func (ip IP) StringExpanded() string {
+	// Invalid and IPv4 are handled the same as in String.
+	switch ip.z {
+	case z0, z4:
+		return ip.String()
+	}
+
+	// See the comment in IP.string6.
+	const max = len("ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff%enp5s0")
```
bradfitz, line 759: why guess on this length? we know the exact length this will be, as we know the Zone (pull it up from the bottom)

netaddr.go
```diff
@@ -653,19 +653,31 @@ func appendDecimal(b []byte, x uint8) []byte {
...
+	noPadHex = false
+)
+
+// appendHex appends the hex string representation of x to b and optionally
+// applies leading zero padding.
+func appendHex(b []byte, x uint16, pad bool) []byte {
 	// Using this function rather than strconv.AppendUint makes IPv6
 	// string building 2x faster.
-
 	if x >= 0x1000 {
```
josharian, line 667: It looks to me like 

```go
if x >= 0x1000 {
 	b = append(b, digits[x>>12])
 } else if pad {
 	b = append(b, '0')
}
```

is equivalent to:

```go
if x >= 0x1000 || pad {
 	b = append(b, digits[x>>12])
}
```

because `digits[0]` is '0'.

And if you rewrite this whole function that way, it starts to look like we may as well have a separate padded version from unpadded version. The padded version is a single call to append with four digits lookups in it. And then we don't have a bool param to kick around.

mdlayher:
Updated!

netaddr.go
```diff
@@ -720,6 +720,14 @@ func appendHex(b []byte, x uint16) []byte {
 	return append(b, digits[x&0xf])
 }
 
+// appendHexPad appends the fully padded hex string representation of x to b.
+func appendHexPad(b []byte, x uint16) []byte {
+	b = append(b, digits[x>>12])
+	b = append(b, digits[x>>8&0xf])
+	b = append(b, digits[x>>4&0xf])
+	return append(b, digits[x&0xf])
```
josharian, line 728: I'd do this all in a single line. It's almost as readable, and it'll generate much better code.

josharian (commented):
Code LGTM, although I'd love an added fuzz test. Something like: For any given IP we have, generate the expanded version and then make sure it parses back to the original IP.

I can't tell from the conversation in the issue whether there's consensus that we should add this API at all.

mdlayher:
Fair enough, I hadn't followed the recent discussion in there. I guess I can leave this alone for now until there is stronger consensus.

bradfitz (commented):
I'm still cool with this.

danderson:
I'm nonplussed, but not opposed. I think so far everyone is either +0 or +1 with no -1, so let's get this reviewed and merged so we can move on :)

netaddr.go
danderson: @mdlayher pls

netaddr.go
```diff
@@ -780,6 +788,43 @@ func (ip IP) string6() string {
 	return string(ret)
 }
 
+// StringExpanded returns the string form of the IP address ip, but IPv6
```
danderson, line 791: Wording nitpicking: I'd write this as "StringExpanded is like String, but <explanation of the IPv6 behavior>". Then there's no need to speak of the v4 and invalid forms being the same, the text makes it clear that everything else is identical to String.

netaddr.go
```diff
@@ -780,6 +788,43 @@ func (ip IP) string6() string {
...
+// addresses are expanded with leading zeroes and no "::" compression. For
+// example, "2001:db8::1" becomes "2001:0db8:0000:0000:0000:0000:0000:0001".
+//
+// IPv4 and invalid IP addresses are formatted the same as when using IP.String.
+// See the documentation of IP.String for more details.
+func (ip IP) StringExpanded() string {
+	// Invalid and IPv4 are handled the same as in String.
```
danderson, line 798: No need for this comment, it's evident enough from the code.

netaddr.go
```diff
@@ -780,6 +788,43 @@ func (ip IP) string6() string {
...
+	switch ip.z {
+	case z0, z4:
+		return ip.String()
+	}
+
+	// Allocate enough space for the longest possible IPv6 address plus the
```
danderson, line 804: This is no longer the longest possible IPv6 address, it's the exact size of the output. Rename variable to `sz` and remove this comment block, once the variable name references size it's sufficiently evident what's going on.

netaddr.go
```diff
@@ -780,6 +788,43 @@ func (ip IP) string6() string {
...
+	// exact length of the zone, if applicable.
+	max := len("ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")
+	if ip.z != z6noz {
+		// Add space for % too.
+		max += 1 + len(ip.Zone())
+	}
+
+	ret := make([]byte, 0, max)
+	for i := uint8(0); i < 8; i++ {
```
danderson, line 813: It feels tempting to unroll this loop for clarity, so that as a reader I don't have to parse out the `i > 0` special case and work things out. Probably not worth it though unless it leads to a massive inlining benefit or somesuch.

(i.e. no action required here, just thinking aloud)

netaddr_test.go
```diff
@@ -760,6 +760,41 @@ func TestLessCompare(t *testing.T) {
 	}
 }
 
+func TestIPStringExpanded(t *testing.T) {
```
danderson, line 763: Benchmark? Do we bench String currently? No need to add a bench if no, but if we test String's perf we should do the same with StringExpanded.

danderson (approved):

netaddr.go
mdlayher: Will leave alone for now!

netaddr_test.go
mdlayher: Added.

mdlayher:
```
09:10 $ go test -bench=BenchmarkIPStringExpanded
goos: linux
goarch: amd64
pkg: inet.af/netaddr
cpu: AMD Ryzen 9 3900X 12-Core Processor            
BenchmarkIPStringExpanded/v4-24         24380799                49.28 ns/op           16 B/op          1 allocs/op
BenchmarkIPStringExpanded/v6-24         10033432               115.7 ns/op            96 B/op          2 allocs/op
BenchmarkIPStringExpanded/v6_ellipsis-24                 9538432               108.3 ns/op            96 B/op          2 allocs/op
BenchmarkIPStringExpanded/v6_v4-24                      10099533               114.4 ns/op            96 B/op          2 allocs/op
BenchmarkIPStringExpanded/v6_zone-24                     9646759               118.7 ns/op            96 B/op          2 allocs/op
PASS
ok      inet.af/netaddr 6.336s
```

Apologies for the delay, I'm freshly back from a week of vacation and not quite firing on all cylinders yet!

mdlayher:
I wonder if we can shave the extra allocation with a strings.Builder or similar:
```
./netaddr.go:849:13: make([]byte, 0, size) escapes to heap:                                                                                                                                        
./netaddr.go:849:13:   flow: {heap} = &{storage for make([]byte, 0, size)}:                                                                                                                        
./netaddr.go:849:13:     from make([]byte, 0, size) (non-constant size) at ./netaddr.go:849:13                                                                                                     
./netaddr.go:862:15: string(ret) escapes to heap:                                                                                                                                                  
./netaddr.go:862:15:   flow: ~r0 = &{storage for string(ret)}:                                                                                                                                     
./netaddr.go:862:15:     from string(ret) (spill) at ./netaddr.go:862:15                                                                                                                           
./netaddr.go:862:15:     from return string(ret) (return) at ./netaddr.go:862:2
```

netaddr.go
```diff
@@ -826,6 +831,37 @@ func (ip IP) appendTo6(ret []byte) []byte {
...
+// zeroes and no "::" compression. For example, "2001:db8::1" becomes
+// "2001:0db8:0000:0000:0000:0000:0000:0001".
+func (ip IP) StringExpanded() string {
+	switch ip.z {
+	case z0, z4:
+		return ip.String()
+	}
+
+	size := len("ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")
+	if ip.z != z6noz {
```
josharian, line 844: Try deleting this if statement entirely. When there are zones, it'll allocate twice, but when there aren't, I think it should cut this to a single allocation. (Constant-sized make can go on the stack.)

netaddr.go
josharian: Another option is to do appendHexPad then append : always, and then slice off the trailing : at the end. Not sure whether that is much clearer (or faster).

netaddr.go
mdlayher:
```
10:02 $ go test -bench=BenchmarkIPStringExpanded 
goos: linux
goarch: amd64
pkg: inet.af/netaddr
cpu: AMD Ryzen 9 3900X 12-Core Processor            
BenchmarkIPStringExpanded/v4-24         28673600                48.45 ns/op           16 B/op          1 allocs/op
BenchmarkIPStringExpanded/v6-24         13939674                83.73 ns/op           48 B/op          1 allocs/op
BenchmarkIPStringExpanded/v6_ellipsis-24                13548344                81.87 ns/op           48 B/op          1 allocs/op
BenchmarkIPStringExpanded/v6_v4-24                      13493323                86.92 ns/op           48 B/op          1 allocs/op
BenchmarkIPStringExpanded/v6_zone-24                     8837446               138.5 ns/op           128 B/op          2 allocs/op
PASS
ok      inet.af/netaddr 6.789s
```

@josharian yep that did the trick!

josharian (approved):

mdlayher:
Merging as-is, we can always keep iterating. Thanks!

PR #140, opened 2021-02-20
ipset: split into a mutable builder and an immutable set.

### LINKED ISSUE #138: Proposal: make IPSet immutable, add IPSetBuilder
Trying to use IPSet in code, it's annoying that it's mutable. It means I have to either defensively copy, or convert into Ranges and reimplement Contains.

Separately, IPSet.ContainsFunc has a todo about maybe using a more efficient representation for pure matching, once the IPSet is constructed.

Empirically, so far all my uses of IPSet have a "build" phase where I add/remove prefixes, followed by a "use" phase where I want read-only use.

How about making a mutable IPSetBuilder, which produces an immutable IPSet? Then we have the freedom to vary IPSet's representation for efficient read ops; IPSet can behave more like the other types in netaddr (no risk of surprise mutations); and we're more explicit about isolating the more expensive operations into a "build" phase. We can have IPSet.ToBuilder(), for the cases where you have an IPSet you like but just want to tweak a bit.

/cc @bradfitz @josharian @mdlayher thoughts?

danderson:
Fixes #138.

### DIFF
--- a/ipset.go
+++ b/ipset.go
@@ -6,15 +6,15 @@ package netaddr
 
 import "sort"
 
-// IPSet represents a set of IP addresses.
+// IPSetBuilder builds an immutable IPSet.
 //
 // The zero value is a valid value representing a set of no IPs.
 //
 // The Add and Remove methods add or remove IPs to/from the set.
-// Add methods should be called first, as a remove operation does
-// nothing on an empty set. Ranges may be fully, partially, or not
-// overlapping.
-type IPSet struct {
+// Removals only affect the current membership of the set, so in
+// general Adds should be called first. Input ranges may overlap in
+// any way.
+type IPSetBuilder struct {
 	// in are the ranges in the set.
 	in []IPRange
 
@@ -22,149 +22,20 @@ type IPSet struct {
 	out []IPRange
 }
 
-// toInOnly updates s to clear s.out, by merging any s.out into s.in.
-func (s *IPSet) toInOnly() {
-	if len(s.out) > 0 {
-		s.in = s.Ranges()
-		s.out = nil
-	}
-}
-
-// Clone returns a copy of s that shares no memory with s.
-func (s *IPSet) Clone() *IPSet {
-	return &IPSet{
-		in: s.Ranges(),
-	}
-}
-
-// Add adds ip to the set s.
-func (s *IPSet) Add(ip IP) { s.AddRange(IPRange{ip, ip}) }
-
-// AddPrefix adds p's range to s.
-func (s *IPSet) AddPrefix(p IPPrefix) { s.AddRange(p.Range()) }
-
-// AddRange adds r to s.
-func (s *IPSet) AddRange(r IPRange) {
-	if !r.Valid() {
-		return
-	}
-	// If there are any removals (s.out), then we need to compact the set
-	// first to get the order right.
-	s.toInOnly()
-	s.in = append(s.in, r)
-}
-
-// Remove removes ip from the set s.
-func (s *IPSet) Remove(ip IP) { s.RemoveRange(IPRange{ip, ip}) }
-
-// RemoveFreePrefix removes and returns a Prefix of length bits from the IPSet.
-func (s *IPSet) RemoveFreePrefix(bitLen uint8) (p IPPrefix, ok bool) {
-	prefixes := s.Prefixes()
-	if len(prefixes) == 0 {
-		return IPPrefix{}, false
-	}
-
-	var bestFit IPPrefix
-	for _, prefix := range prefixes {
-		if prefix.Bits > bitLen {
-			continue
-		}
-		if bestFit.IP.IsZero() || prefix.Bits > bestFit.Bits {
-			bestFit = prefix
-			if bestFit.Bits == bitLen {
-				// exact match, done.
-				break
-			}
-		}
-	}
-
-	if bestFit.IP.IsZero() {
-		return IPPrefix{}, false
-	}
-
-	prefix := IPPrefix{IP: bestFit.IP, Bits: bitLen}
-	s.RemovePrefix(prefix)
-	return prefix, true
-}
-
-// RemovePrefix removes p's range from s.
-func (s *IPSet) RemovePrefix(p IPPrefix) { s.RemoveRange(p.Range()) }
-
-// RemoveRange removes r from s.
-func (s *IPSet) RemoveRange(r IPRange) {
-	if r.Valid() {
-		s.out = append(s.out, r)
-	}
-}
-
-// AddSet adds all ranges in b to s.
-func (s *IPSet) AddSet(b *IPSet) {
-	for _, r := range b.Ranges() {
-		s.AddRange(r)
-	}
-}
-
-// RemoveSet removes all ranges in b from s.
-func (s *IPSet) RemoveSet(b *IPSet) {
-	for _, r := range b.Ranges() {
-		s.RemoveRange(r)
-	}
-}
-
-// Complement updates s to contain the complement of its current
-// contents.
-func (s *IPSet) Complement() {
-	s.toInOnly()
-	s.out = s.in
-	s.in = []IPRange{
-		IPPrefix{IP: IPv4(0, 0, 0, 0), Bits: 0}.Range(),
-		IPPrefix{IP: IPv6Unspecified(), Bits: 0}.Range(),
-	}
-}
-
-// Intersect updates s to the set intersection of s and b.
-func (s *IPSet) Intersect(b *IPSet) {
-	b = b.Clone()
-	b.Complement()
-	s.RemoveSet(b)
-}
-
-// Overlaps returns whether s and b have any IPs in common.
-func (s *IPSet) Overlaps(b *IPSet) bool {
-	sr, br := s.Ranges(), b.Ranges()
-	for len(sr) > 0 && len(br) > 0 {
-		sr0, br0 := sr[0], br[0]
-		switch {
-		case sr0.Overlaps(br0):
-			return true
-		case sr0.entirelyBefore(br0):
-			sr = sr[1:]
-		default:
-			br = br[1:]
-		}
-	}
-	return false
-}
-
-func discardf(format string, args ...interface{}) {}
-
-// debugf is reassigned by tests.
-var debugf = discardf
-
-// Ranges returns the minimum and sorted set of IP
-// ranges that covers s.
-func (s *IPSet) Ranges() []IPRange {
+// normalize normalizes s: s.in becomes the minimal sorted list of
+// ranges required to describe s, and s.out becomes empty.
+func (s *IPSetBuilder) normalize() {
 	const debug = false
 	if debug {
 		debugf("ranges start in=%v out=%v", s.in, s.out)
 	}
 	in, ok := mergeIPRanges(s.in)
 	if !ok {
-		return nil
+		return
 	}
 	out, ok := mergeIPRanges(s.out)
 	if !ok {
-		return nil
+		return
 	}
 	if debug {
 		debugf("ranges sort  in=%v out=%v", in, out)
@@ -174,7 +45,7 @@ func (s *IPSet) Ranges() []IPRange {
 	// overlaps within each other. We can run a merge of the two lists
 	// in one pass.
 
-	ret := make([]IPRange, 0, len(in))
+	min := make([]IPRange, 0, len(in))
 	for len(in) > 0 && len(out) > 0 {
 		rin, rout := in[0], out[0]
 		if debug {
@@ -200,11 +71,11 @@ func (s *IPSet) Ranges() []IPRange {
 			//
 			//    in         out
 			// f------t   f-------t
-			ret = append(ret, rin)
+			min = append(min, rin)
 			in = in[1:]
 			if debug {
 				debugf("in before out; append in")
-				debugf("ret=%v", ret)
+				debugf("min=%v", min)
 			}
 		case rin.coveredBy(rout):
 			// "out" entirely covers "in".
@@ -224,14 +95,14 @@ func (s *IPSet) Ranges() []IPRange {
 			// f-------------t
 			//    f------t
 			//       out
-			ret = append(ret, IPRange{From: rin.From, To: rout.From.Prior()})
+			min = append(min, IPRange{From: rin.From, To: rout.From.Prior()})
 			// Adjust in[0], not ir, because we want to consider the
 			// mutated range on the next iteration.
 			in[0].From = rout.To.Next()
 			out = out[1:]
 			if debug {
 				debugf("out inside in; split in, append first in, drop out, adjust second in")
-				debugf("ret=%v", ret)
+				debugf("min=%v", min)
 			}
 		case rout.overlapsStartOf(rin):
 			// "out" overlaps start of "in".
@@ -241,7 +112,7 @@ func (s *IPSet) Ranges() []IPRange {
 			//    f------t
 			//       in
 			in[0].From = rout.To.Next()
-			// Can't move ir onto ret yet, another later out might
+			// Can't move ir onto min yet, another later out might
 			// trim it further. Just discard or and continue.
 			out = out[1:]
 			if debug {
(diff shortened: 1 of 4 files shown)


ipset.go
```diff
@@ -4,167 +4,36 @@
 
 package netaddr
 
-import "sort"
-
-// IPSet represents a set of IP addresses.
+// IPSetBuilder is an IPSet builder.
```
bradfitz, line 7: IPSetBuilder builds an immutable IPSet.

?

Less tautological?

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+}
+
+// Add adds ip to the set s.
+func (s *IPSetBuilder) Add(ip IP) { s.AddRange(IPRange{ip, ip}) }
+
+// AddPrefix adds p's range to s.
+func (s *IPSetBuilder) AddPrefix(p IPPrefix) { s.AddRange(p.Range()) }
 
-	return ret
+// AddRange adds r to s.
```
bradfitz, line 156: If r is not Valid, AddRange does nothing.

?

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	o.RemoveSet(b)
+	s.removeBuilder(&o)
+}
+
+func discardf(format string, args ...interface{}) {}
+
+// debugf is reassigned by tests.
+var debugf = discardf
+
+func (s *IPSetBuilder) IPSet() *IPSet {
```
bradfitz, line 257: doc?

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	return true
 }
 
-// Equal reports whether s contains exactly the same IPs as t.
-func (s *IPSet) Equal(t *IPSet) bool {
-	sr := s.Ranges()
-	tr := t.Ranges()
-	if len(sr) != len(tr) {
-		return false
+// Contains returns whether ip is in s.
```
bradfitz, line 308: reports

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSet) Contains(ip IP) bool {
+	for _, r := range s.rr {
```
bradfitz, line 310: // TODO use a better data structure with link to that github issue?

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) IPSet() *IPSet {
+	s.normalize()
+	return &IPSet{
+		rr: append([]IPRange{}, s.in...),
+	}
+}
+
+// IPSet represents a set of IP addresses.
+//
+// The zero value is a valid value representing a set of no IPs.
```
bradfitz, line 266: Add that it's safe for concurrent use.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+type IPSet struct {
+	rr []IPRange // Sorted by IPRange.less
+}
+
+// Clone returns a copy of s that shares no memory with s.
+func (s *IPSet) Clone() *IPSet {
```
bradfitz, line 272: But why if it's immutable?

Oh, mention that it's immutable in the type's godoc too?

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+			return true
+		}
 	}
-	for i := range sr {
-		if sr[i] != tr[i] {
-			return false
+	return false
+}
+
+// ContainsRange returns whether all IPs in o are in s.
```
bradfitz, line 318: "reports whether" here and elsewhere

ipset.go
```diff
@@ -4,167 +4,36 @@
...
-// The Add and Remove methods add or remove IPs to/from the set.
-// Add methods should be called first, as a remove operation does
-// nothing on an empty set. Ranges may be fully, partially, or not
-// overlapping.
-type IPSet struct {
+// The Add and Remove methods add or remove IPs to/from the
+// set. Removals only affect the current membership of the set, so in
+// general Adds should be called first. Input ranges may overlap in
+// any way.
+type IPSetBuilder struct {
```
bradfitz, line 15: regarding your name qualms: if this were called MutableIPSet, that would make it obvious that IPSet weren't mutable, and would make `RemoveFreePrefix` less gross (not being a method on a "builder" type)

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+			debugf("min=%v", min)
 		}
 	}
 
-	// TODO: possibly update s.in and s.out, if #110 supports that.
+	s.in = min
+	s.out = nil
+}
+
+// Add adds ip to the set s.
```
josharian, line 150: s/the set //

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) Add(ip IP) { s.AddRange(IPRange{ip, ip}) }
+
+// AddPrefix adds p's range to s.
```
josharian, line 153: Or maybe clearer

// AddPrefix adds all IPs in p to s.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	}
+	// If there are any removals (s.out), then we need to compact the set
+	// first to get the order right.
+	if len(s.out) > 0 {
+		s.normalize()
+	}
+	s.in = append(s.in, r)
+}
+
+// AddSet adds all ranges in b to s.
```
josharian, line 169: s/all ranges/all IPs/

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) AddSet(b *IPSet) {
+	for _, r := range b.rr {
+		s.AddRange(r)
+	}
+}
+
+// Remove removes ip from the set s.
```
josharian, line 176: s/the set //

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) Remove(ip IP) { s.RemoveRange(IPRange{ip, ip}) }
+
+// RemoveFreePrefix removes and returns a Prefix of length bits from the IPSetBuilder.
```
josharian, line 179: See my API comment on the corresponding issue.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	if bestFit.IP.IsZero() {
+		return IPPrefix{}, false
+	}
+
+	prefix := IPPrefix{IP: bestFit.IP, Bits: bitLen}
+	s.RemovePrefix(prefix)
+	return prefix, true
+}
+
+// RemovePrefix removes p's range from s.
```
josharian, line 208: // RemovePrefix removes all IPs in p from s.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) RemovePrefix(p IPPrefix) { s.RemoveRange(p.Range()) }
+
+// RemoveRange removes r from s.
```
josharian, line 211: // RemoveRange removes all IPs in r from s.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSetBuilder) RemoveRange(r IPRange) {
+	if r.Valid() {
+		s.out = append(s.out, r)
+	}
+}
+
+// RemoveSet removes all ranges in o from s.
```
josharian, line 218: // RemoveSet removes all IPs in o from s.

Also, why `o` here and `b` above? Let's stick with `b`, because it's less visually ambiguous.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+
+// removeBuilder removes all IPs in b from s.
+func (s *IPSetBuilder) removeBuilder(b *IPSetBuilder) {
+	b.normalize()
+	for _, r := range b.in {
+		s.RemoveRange(r)
+	}
+}
+
+// Complement updates s to contain the complement of its current
```
josharian, line 233: I thought we were going to add a universe argument to complement? Or at least document what the universe is.

ipset.go
josharian: Add

// Use IPSetBuilder to construct IPSets.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	return &IPSet{
+		rr: append([]IPRange{}, s.in...),
+	}
+}
+
+// IPSet represents a set of IP addresses.
+//
+// The zero value is a valid value representing a set of no IPs.
+type IPSet struct {
+	rr []IPRange // Sorted by IPRange.less
```
josharian, line 268: nit: s/Sorted/sorted/

ipset.go
josharian: I can see wanting an IPSetBuilder `Clone` method.

ipset.go
josharian: Can we use `r` for consistency? We can pick a different local variable name (sr? x?); this one is more important.

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	for _, r := range s.rr {
+		if o.coveredBy(r) {
+			return true
 		}
 	}
-	return true
+	return false
+}
+
+// ContainsPrefix returns whether all IPs in p are in s.
```
josharian, line 328: s/returns/reports/

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+func (s *IPSet) ContainsPrefix(p IPPrefix) bool {
+	return s.ContainsRange(p.Range())
+}
+
+// Overlaps returns whether any IP in o is also in s.
```
josharian, line 333: s/returns/reports/

josharian: And s/o/b/ (or t if you don't like b)

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+		for _, or := range o.rr {
+			if r.Overlaps(or) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// OverlapsRange returns whether any IP in o is also in s.
```
josharian, line 346: reports
t or b

ipset.go
```diff
@@ -268,59 +137,224 @@ func (s *IPSet) Ranges() []IPRange {
...
+	for _, r := range s.rr {
+		if r.Overlaps(o) {
+			return true
+		}
+	}
+	return false
+}
+
+// OverlapsPrefix returns whether any IP in p is also in s.
+func (s *IPSet) OverlapsPrefix(p IPPrefix) bool {
```
josharian, line 358: Is it worth having all all these API variants, when it's so easy to get a range out of a prefix? (Maybe it is, but I want to make sure we consider the question.)

josharian (commented):
(My GitHub notifications for netaddr go to my Tailscale email address, so expect slow turnarounds from me on any netaddr discussions...unless explicitly mentioned, in which case my phone dings. 🤷🏽 )

ipset.go
danderson: Reminds me I want to maybe change that, the whole "mysteriously do nothing on invalid input" makes me a bit nervous. Filing bug about that.

ipset.go
danderson: Good point, no reason to have Clone here. Moved to IPSetBuilder.

ipset.go
danderson: MutableIPSet suggests that it should have all the various accessors that IPSet has, in my mind. I'd rather avoid that and keep it as a pure builder type.

I'm leaning a little more to Josh's suggestion of moving RemoveFreePrefix over to IPSet, and just make it produce a new IPSet on use. Makes it a little more expensive to use, but it's not a hot path.

ipset.go
danderson: Separate issue. Going to leave this as-is for now, as it's as broken as the previous code.

ipset.go
danderson: Used r, for consistency with other IPRange methods.

ipset.go
danderson: I didn't consider it too hard, honestly. I figured it's cheap for us to provide accessors for all the top-level types in netaddr, rather than require callers to add visual noise.

ipset.go
danderson: Moved RemoveFreePrefix onto IPSet.

ipset.go
```diff
@@ -268,59 +137,233 @@ func (s *IPSet) Ranges() []IPRange {
...
-	return true
+	return false
+}
+
+// ContainsPrefix reports whether all IPs in p are in s.
+func (s *IPSet) ContainsPrefix(p IPPrefix) bool {
+	return s.ContainsRange(p.Range())
+}
+
+// Overlaps reports whether any IP in o is also in s.
```
bradfitz, line 311: `"in o"` but the arguments are named `s` and `b.

ipset.go
```diff
@@ -268,59 +137,233 @@ func (s *IPSet) Ranges() []IPRange {
...
+		for _, or := range b.rr {
+			if r.Overlaps(or) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// OverlapsRange reports whether any IP in o is also in s.
```
bradfitz, line 324: same

ipset.go
```diff
@@ -268,59 +137,233 @@ func (s *IPSet) Ranges() []IPRange {
...
+	return false
+}
+
+// OverlapsPrefix reports whether any IP in p is also in s.
+func (s *IPSet) OverlapsPrefix(p IPPrefix) bool {
+	return s.OverlapsRange(p.Range())
+}
+
+// RemoveFreePrefix splits s into a Prefix of length bitLen and a new
+// IPSet with that prefix removed.
```
bradfitz, line 341: worth mentioning what ok==false means? That no such space exists.

ipset.go
```diff
@@ -268,59 +137,233 @@ func (s *IPSet) Ranges() []IPRange {
...
+	}
+}
+
+// Add adds ip to s.
+func (s *IPSetBuilder) Add(ip IP) { s.AddRange(IPRange{ip, ip}) }
+
+// AddPrefix adds all IPs in p to s.
+func (s *IPSetBuilder) AddPrefix(p IPPrefix) { s.AddRange(p.Range()) }
+
+// AddRange adds r to s. If r is not Valid, AddRange does nothing.
```
bradfitz, line 164: If on a new line

Also should the other two Add methods above also say the equivalent thing?

danderson: The other Add methods take types which cannot be invalid, so no such problem applies.

ipset.go
```diff
@@ -268,59 +137,237 @@ func (s *IPSet) Ranges() []IPRange {
...
-	// with s.Prefixes()?
-	return func(ip IP) bool {
-		i := sort.Search(len(rv), func(i int) bool {
-			return ip.Less(rv[i].From)
-		})
-		if i == 0 {
+// Equal reports whether s and o represent the same set of IP
+// addresses.
+func (s *IPSet) Equal(o *IPSet) bool {
+	if len(s.rr) != len(o.rr) {
```
bradfitz, line 274: this implementation assumes that s.rr is also normalized; add a comment to line 252 where it says "sorted by IPRange.less" to say a bit more? (might need to promote it to a not-same-line comment)

ipset.go
```diff
@@ -268,59 +137,237 @@ func (s *IPSet) Ranges() []IPRange {
...
 
-// ContainsFunc returns a func that reports whether an IP is in s.
-// The returned func operates on a copy of s, so s may be mutated
-// later.
-func (s *IPSet) ContainsFunc() (contains func(IP) bool) {
-	rv := s.Ranges()
-	// TODO(bradfitz): build a faster data structure with
-	// with s.Prefixes()?
-	return func(ip IP) bool {
-		i := sort.Search(len(rv), func(i int) bool {
```
bradfitz, line 302: why did you delete this? This was a better implementation of Contains, no?

bradfitz (approved):

ipset.go
danderson: Documented.

ipset.go
danderson: I deleted ContainsFunc because we now have a plain Contains. Not porting the implementation over was an oversight. Fixed.

PR #142, opened 2021-02-23
netaddr: create IPPrefix from netmask

### LINKED ISSUE #133: IPPrefix from Mask
Right now there are four ways to construct an `IPPrefix`:

1. Parsing an string with an IP represented in CIDR notation: `netaddr.ParseIPPrefix("127.0.0.1/8")`
2. Providing the significant bits to an `IP` struct: `netaddr.IPv4(127, 0, 0, 1).Prefix(8)`
3. Working with `IPRange`: `netaddr.ParseRange("127.0.0.0-127.0.0.255").Prefix()` (ignoring error returns)
4. Working with `IPSet`: `ipset.Prefixes()`

I'm working with a kernel API that returns the IP mask itself[^1]. I can turn this into an IPPrefix by first using `net.IPMask` and calling `Size`.

```go
func main() {
	var mask [4]byte
	netmask := make(net.IPMask, net.IPv4len)

	mask = getKernelMask()
	copy(netmask, mask[:])

	bits, _ := netmask.Size()
	ip, _ := netaddr.IPv4(127, 0, 0, 1).Prefix(uint8(bits))

	fmt.Printf("IPPrefix: %s\n", ip)
}
```

Would there be any interest in allowing an `IPPrefix` to be constructed either directly from a netmask representation, or alternatively from a `net.IPMask`?

[^1]: The kernel API returns the netmask for IPv4. For IPv6 they switched to returning the significant bits instead.

terinjokes:
Creates an IPPrefix from an IP and a netmask byte slice.

Fixes #133

Signed-off-by: Terin Stock <terinjokes@gmail.com>

### DIFF
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -829,6 +829,174 @@ func TestIPPrefixMasking(t *testing.T) {
 	}
 }
 
+func TestIPNetmasking(t *testing.T) {
+	type subtest struct {
+		ip   IP
+		mask []byte
+		p    IPPrefix
+		ok   bool
+	}
+
+	// makeIPv6 produces a set of IPv6 subtests with an optional zone identifier.
+	makeIPv6 := func(zone string) []subtest {
+		if zone != "" {
+			zone = "%" + zone
+		}
+
+		return []subtest{
+			{
+				ip: mustIP(fmt.Sprintf("2001:db8::1%s", zone)),
+				mask: []byte{
+					0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+					0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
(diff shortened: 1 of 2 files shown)


netaddr_test.go
```diff
@@ -829,6 +829,157 @@ func TestIPPrefixMasking(t *testing.T) {
...
+
+					p, err := st.ip.Netmask(st.mask)
+					if st.ok && err != nil {
+						t.Fatalf("failed to produce prefix: %v", err)
+					}
+					if !st.ok && err == nil {
+						t.Fatal("expected an error, but none occurred")
+					}
+
+					if err !=nil {
```
mdlayher, line 965: Needs gofmt.

mdlayher (commented):
LGTM overall. Please add test cases for the non-contiguous subnet mask check too.

Will leave for others to ensure consensus.

mdlayher (approved):
LGTM.

mdlayher:
Since there have been no objections, I'm going to merge this since it seems like a useful API to land. We can always make adjustments before we tag a v1.0.0.

PR #148, opened 2021-03-12
Add IP.{Interface,Link}LocalMulticast methods

tklauser:
Corresponding to the respective standard library `net.IP` methods: [`net.IP.IsInterfaceLocalMulticast`](https://golang.org/pkg/net/#IP.IsInterfaceLocalMulticast) and [`net.IP.IsLinkLocalMulticast`](https://golang.org/pkg/net/#IP.IsLinkLocalMulticast).

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -564,6 +564,16 @@ func (ip IP) IsMulticast() bool {
 	return false // zero value
 }
 
+// IsInterfaceLocalMulticast reports whether ip is an IPv6 interface-local
+// multicast address. If ip is the zero value or an IPv4 address, it will return
+// false.
+func (ip IP) IsInterfaceLocalMulticast() bool {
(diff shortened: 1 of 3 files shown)


netaddr.go
```diff
@@ -564,6 +564,27 @@ func (ip IP) IsMulticast() bool {
 	return false // zero value
 }
 
+// IsInterfaceLocalMulticast reports whether ip is an interface-local multicast
```
mdlayher, line 567: Seems worth noting that this only applies to IPv6 addresses in the doc comment.

mdlayher (approved):
LGTM with a docs nit.

netaddr.go
tklauser: Done, thanks.

mdlayher (approved):
LGTM thanks

PR #149, opened 2021-03-13
RFC: IPPrefix.String() should return the normalized network

majst01:
Hi,

i stumbled about the following:

```
pfx1, err := ParseIPPrefix("10.0.0.0/24") 
pfx1.String() // 10.0.0.0/24

pfx2, err := ParseIPPrefix("10.0.0.1/24") 
pfx2.String() // 10.0.0.1/24
```
From my understanding both are the same Prefix so the Stringer should return the same result in both cases.
Same applies for IPv6.

Modified TestIPPrefix with a dump implementation of IPPrefix.String() just for reference.

WDYT ?

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -1143,7 +1143,9 @@ func (p IPPrefix) String() string {
 	if !p.Valid() {
 		return "invalid IP prefix"
 	}
-	return fmt.Sprintf("%s/%d", p.IP, p.Bits)
+	// TODO remore stdlib impl
+	network := p.IP.IPAddr().IP.Mask(p.IPNet().Mask)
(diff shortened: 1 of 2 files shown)


mdlayher:
I forget the exact details but we chose to add a Masked method for actually applying the mask to turn the IPPrefix into its canonical form: https://pkg.go.dev/inet.af/netaddr?utm_source=godoc#IPPrefix.Masked

For that reason I think we probably want to leave this as-is and make it the caller's responsibility to call Masked to return the canonical form. But that's just my two cents.

majst01:
Perfect thanks for clarification.

mdlayher:
No worries!

PR #150, opened 2021-03-15
Run go mod tidy as part of GitHub actions

tklauser:
Make sure `go.mod` and `go.sum` match the current source in the module.

Running `go mod tidy` also removes the indirect dependency on
github.com/dvyukov/go-fuzz which was probably added when running the `go
get` commands mentioned in the README to install the `go-fuzz` and
`go-fuzz-build` in the package directory.


go.mod
```diff
@@ -3,7 +3,6 @@ module inet.af/netaddr
 go 1.12
 
 require (
-	github.com/dvyukov/go-fuzz v0.0.0-20201127111758-49e582c6c23d // indirect
```
bradfitz, line 6: Hm, how do you feel about this @josharian?

Seems fine (actually even good) to me, but is there any reason you wanted this here? If you need to pin a version, do it in a shell script or Makefile that runs the fuzz?

josharian: Every time you run go fuzz, it adds that dependency. It's long story...probably fixable, but not worth the effort. So I checked it in so that I could run go fuzz without dirtying the repo. I'd rather leave it in if possible. Or maybe there's an easy way to make go mod tidy leave it there?

tklauser: What I've seen other projects do in the same situation is adding a separate package (e.g. `tools`) importing the dependency. This will then show up in `go.mod`, i.e. for this repo something along the lines of:

```
$ mkdir tools
$ cat > tools/tools.go << EOF
// +build tools

package tools

import (
	_ "github.com/dvyukov/go-fuzz/go-fuzz-build"
)
EOF
$ go mod tidy
```

Does that sound reasonable to both of you @josharian @bradfitz?

josharian: Fine by me.

tklauser: Pushed an additional commit adding the above.

bradfitz (approved):

PR #163, opened 2021-05-14
make types opaque

### LINKED ISSUE #154: make IPPrefix opaque, smaller?
`IPPrefix` is currently defined as:

```go
type IPPrefix struct {
	IP   IP
	Bits uint8
}
```

That means it takes 32 bytes (25 bytes + padding: https://play.golang.org/p/4XZ9LQWEwj4).

And it also means callers need to create them with keyed struct literals, lest they anger `go vet`'s cross-package-unkeyed-struct-literal check:

     ipp := netaddr.IPPrefix{IP: ..., Bits: ....}

We extolled the virtues of opaque representations in https://tailscale.com/blog/netaddr-new-ip-type-for-go/, but our IPPrefix type isn't.

We could instead do:

```go
type IPPrefix struct {
     addr uint128
     isv6 bool
     bits uint8
}
```

And we'd be at 24 bytes after padding: https://play.golang.org/p/oS6tg8RtycQ

We'd then need methods:

```go
func (p IPPrefix) IP() IP
func (p IPPrefix) Bits() uint8
```

But those would be alloc-free.

We'd then need a constructor. I can't think of a great. `NewFoo` should return a _pointer_ to `Foo`, so `NewIPPrefix` isn't great. `Make` isn't great. Maybe:

* `func Prefix(ip IP, bits uint8) IPPrefix` ?
* `func IPBits(ip IP, bits uint8) IPPrefix` ?

/cc @danderson @josharian @mdlayher @crawshaw

josharian:
I plan to make a code rewriter to help users adapt to these changes.
That'll be in a separate repo.

Fixes #154.


netaddr.go
```diff
@@ -859,14 +859,27 @@ func (ip *IP) UnmarshalBinary(b []byte) error {
...
-// It's meant to be used as a value type.
+// IPPort is an IP and a port number.
 type IPPort struct {
-	IP   IP
-	Port uint16
+	ip   IP
+	port uint16
 }
 
+// IPPortFrom returns an IPPort with ip ip and port port.
```
bradfitz, line 868: with IP ip (and elsewhere)

netaddr.go
```diff
@@ -859,14 +859,27 @@ func (ip *IP) UnmarshalBinary(b []byte) error {
...
+func IPPortFrom(ip IP, port uint16) IPPort { return IPPort{ip: ip, port: port} }
+
+// WithIP returns an IPPort with ip ip and port p.Port().
+func (p IPPort) WithIP(ip IP) IPPort { return IPPort{ip: ip, port: p.port} }
+
+// WithIP returns an IPPort with ip p.IP() and port port.
```
bradfitz, line 874: Worth documenting that these don't alloc? Or maybe just in package doc is sufficient if we add it there once?

bradfitz (approved):

PR #168, opened 2021-05-18
add AppendTo method to IP, IPPort, and IPPrefix

josharian:


netaddr.go
```diff
@@ -1263,31 +1288,43 @@ func (p IPPrefix) Overlaps(o IPPrefix) bool {
...
+	return b
+}
+
+// MarshalText implements the encoding.TextMarshaler interface,
+// The encoding is the same as returned by String, with one exception:
+// If p is the zero value, the encoding is the empty string.
+func (p IPPrefix) MarshalText() ([]byte, error) {
+	var max int
+	switch p.ip.z {
+	case z0:
```
danderson, line 1320: Could you return nil here, or does MarshalText mandate returning a non-nil zero length slice?

danderson (approved):

netaddr.go
josharian: Good question. I don't see anything in the docs that require non-nil. I'm going to leave as-is here to avoid changing behavior at the same time as adding API. I'll send a follow-up shortly.

josharian: This turned out to be a really good question. Our marshaling behavior is rather inconsistent and probably incorrect in some cases.

PR #171, opened 2021-05-19
correctly parse IPPrefix "::%0/00/80"

josharian:


bradfitz (approved):

netaddr_test.go
```diff
@@ -1305,6 +1305,17 @@ func TestIPPrefix(t *testing.T) {
 			contains:    mustIPs("2001:db8::1"),
 			notContains: mustIPs("fe80::1"),
 		},
+		{
+			prefix: "::%0/00/80",
+			ip:     mustIP("::%0/00"),
```
danderson, line 1310: That's not a valid result, is it? Are zones allowed to contain slashes?

danderson (changes requested):

netaddr_test.go
mdlayher: I've never seen a zone with a slash. I'm not sure if / is a valid character for network interface names.

josharian: There are no restrictions on what can be in a zone. Quoting [rfc 6874](https://datatracker.ietf.org/doc/html/rfc6874):

> According to IPv6 Scoped Address syntax [RFC4007], a zone identifier is attached to the textual representation of an IPv6 address by concatenating "%" followed by <zone_id>, where <zone_id> is a string identifying the zone of the address.  However, the IPv6 Scoped Address Architecture specification gives no precise definition of the character set allowed in <zone_id>.  There are no rules or de facto standards for this.  For example, the first Ethernet interface in a host might be called %0, %1, %en1, %eth0, or whatever the implementer happened to choose.

I have strong feelings about this, but...it is what it is.

josharian: It's worth noting that IPPort doesn't have this problem, because it uses square brackets around the IP address when using IPv6. On pain of inventing a new syntax, we could do the same thing for prefixes. But that seems unwise, given that this is an extremely rare case; LastIndexByte seems ample.

josharian:
Dave, you OK with me proceeding with this?

mdlayher (approved):
LGTM.

netaddr_test.go
danderson: With regret, LGTM if we're implementing the spec as well as it can be implemented. Hopefully no user will ever try such a cursed address for real.

danderson (approved):

PR #172, opened 2021-05-19
improve handling of invalid IPPorts

josharian:


netaddr.go
```diff
@@ -973,13 +973,20 @@ func MustParseIPPort(s string) IPPort {
 // IsZero reports whether p is its zero value.
 func (p IPPort) IsZero() bool { return p == IPPort{} }
 
+// Valid reports whether p.IP() is non-zero.
```
bradfitz, line 976: Worth explicitly mentioning that the port is ignored, or that a zero port may be valid?

netaddr_test.go
```diff
@@ -2726,3 +2726,19 @@ func TestNoAllocs(t *testing.T) {
...
 	})
 }
+
+func TestIPPortString(t *testing.T) {
+	tests := []struct {
+		ipp  IPPort
+		want string
+	}{
+		{IPPort{}, "invalid IPPort"},
+		{IPPortFrom(IP{}, 80), "invalid IPPort"},
```
bradfitz, line 2736: add a test to lock in behavior of non-zero IP and zero port.

josharian: We have a fair amount of those in TestParseIPPort. I'll name this function to make it clearer that it is about invalid IPPorts.

netaddr.go
josharian: Done.

netaddr.go
```diff
@@ -971,15 +971,23 @@ func MustParseIPPort(s string) IPPort {
 }
 
 // IsZero reports whether p is its zero value.
+// All ports are valid, including zero.
```
danderson, line 974: This comment should go in Valid's docstring or the type's docstring, not IsZero's.

danderson (approved):

netaddr.go
josharian: Oops. Fixed, thanks.

PR #174, opened 2021-05-19
fuzz more invalid values

### LINKED ISSUE #169: IPPort and IPPrefix MarshalText doesn't round trip
I'm puzzling over how IPPort and IPPrefix's MarshalText methods should work when their IP is the zero IP but they have a non-zero port/bits.

Currently we return an empty slice, but that doesn't round trip. If you do `IPPortFrom(IP{}, 80)`, `MarshalText`, and then `UnmarshalText`, you get a zero `IPPort`. That is, we lose the `80`. Similarly so for `IPPrefix`.

Options:

1. Make MarshalText return an error when there is a zero IP and a non-zero port/prefix. We then lose the nice property that MarshalText never returns an error.
2. Make MarshalText return something like `:80` (or `invalid IP:80`?), and then teach UnmarshalText to handle those inputs. We then lose the nice property that UnmarshalText is identical to ParseIPPort except in its handling of an empty input.
3. Do the previous option, and also change ParseIPPort to match UnmarshalText. We then lose the nice property that there is no way to get a zero IP by parsing a textual input, only by constructing one in code.

I'm inclined towards 1, but don't feel strongly. Opinions?

cc @danderson @maisem @bradfitz @mdlayher

josharian:
Two commits. The first is safe to merge any time.
The second one is not: It'll cause a bunch of fuzz failures.
We need to fix #169 first.
The new fuzzing power is pulling its weight:
It found a few other new, unrelated bugs,
which I have sent fixes for.

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -8,8 +8,10 @@ package netaddr
 
 import (
 	"bytes"
+	"encoding"
 	"fmt"
 	"net"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


fuzz.go
```diff
@@ -67,107 +43,155 @@ func Fuzz(b []byte) int {
...
+
+// parseType are trampoline functions that give ParseType functions the same signature.
+// This would be nicer with generics.
+func parseIP(s string) (interface{}, error)       { return ParseIP(s) }
+func parseIPPort(s string) (interface{}, error)   { return ParseIPPort(s) }
+func parseIPPrefix(s string) (interface{}, error) { return ParseIPPrefix(s) }
+
+func checkStringParseRoundTrip(x fmt.Stringer, parse func(string) (interface{}, error)) {
+	// Zero values tend to print something like "invalid IP", so it's OK if they don't round trip.
+	if z, ok := x.(interface{ IsZero() bool }); ok {
```
danderson, line 158: Can you structure the zero and valid checks such that `IsZero() && Valid()` does get checked for round-tripping? We should not care about invalid zero values failing to round-trip, but a valid zero value should.

danderson (approved):

fuzz.go
josharian: IIUC, that is equivalent to only checking `Valid` here. That currently fails: `IP{}.String() == "invalid IP"`, but `ParseIP("invalid IP")` yields an error. (Note that we're talking about the Go zero value, not `0.0.0.0` or `::` or the like.) Or have I perhaps misunderstood?

danderson: Why is IP{} `Valid` ? That seems like an invalid value to me?

Let's lock in the current behavior (i.e. please submit this with no changes needed) and we can figure out if there's a refinement needed separately.

danderson: Ah, I think I see the issue. The check I'm thinking of is: if x implements _both_ Valid and IsZero, and the zero value is Valid, then test it. I think that'd get things working, because `IP` has no Valid implementation, and so would still get skipped. Am I making sense?

josharian: Ah! I understand. Good idea, done.

PR #176, opened 2021-05-21
Make IPRange.less sort subset ranges after their superset.

### LINKED ISSUE #145: IPRange sort order is inconsistent with coverage (superset/subset)
Hi, thanks a lot for netaddr! 
Just my 5 cents: In my opinion is the sort order for IPRange.less() semi optimal:

```
// less returns whether r is "before" other. It is before if r.From is
// before other.From, or if they're equal, the shorter range is
// before.
func (r IPRange) less(other IPRange) bool {
	if cmp := r.From.Compare(other.From); cmp != 0 {
		return cmp < 0
	}
	return r.To.Less(other.To)
}
```
better would be if *all* subsets are sorted *after* their supersets. Think about this small change below:

```
// less returns whether r is "before" other. It is before if r.From is
// before other.From, or if they're equal, r is superset to other.
// If coveredBy is true, less must be false. subsets are always sorted "after" their supersets.
func (r IPRange) less(other IPRange) bool {
	if cmp := r.From.Compare(other.From); cmp != 0 {
		return cmp < 0
	}
       // all supersets to the left, subsets to the right
	return other.To.Less(r.To)
}
```
With the current sort order, the sorting of a superset to a subset is different depending if the To's OR From's are equal. Sometimes before, sometimes after, depending on the range borders and not also on the coverage.

I personally work on an interval tree for ip ranges (not only CIDRs, tries are not possible!), it is important that the supersets are always sorted **on the same side** of their including subsets.

Best regards
  Charly

PS: the tests still pass ;)

danderson:
This doesn't alter the behavior of dependent code, but does
make the sort order slightly more "natural" and consistent.

Fixes #145.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -1465,14 +1465,14 @@ func (r IPRange) contains(addr IP) bool {
 	return r.from.Compare(addr) <= 0 && r.to.Compare(addr) >= 0
 }
 
-// less returns whether r is "before" other. It is before if r.From() is
-// before other.From(), or if they're equal, the shorter range is
-// before.
+// less reports whether r is "before" other. It is before if r.From()
+// is before other.From(). If they're equal, then the larger range
+// (higher To()) comes first.
 func (r IPRange) less(other IPRange) bool {
 	if cmp := r.from.Compare(other.from); cmp != 0 {
 		return cmp < 0
 	}
-	return r.to.Less(other.to)
+	return other.to.Less(r.to)
 }
 
 // entirelyBefore returns whether r lies entirely before other in IP


netaddr.go
```diff
@@ -1465,14 +1465,14 @@ func (r IPRange) contains(addr IP) bool {
 	return r.from.Compare(addr) <= 0 && r.to.Compare(addr) >= 0
 }
 
-// less returns whether r is "before" other. It is before if r.From() is
-// before other.From(), or if they're equal, the shorter range is
-// before.
+// less returns whether r is "before" other. It is before if r.From()
```
josharian, line 1468: Can we use this opportunity to switch to "reports whether"?

netaddr.go
```diff
@@ -1465,14 +1465,14 @@ func (r IPRange) contains(addr IP) bool {
...
+// is before other.From(), or if they're equal, the larger range comes
```
josharian, line 1469: Let's split into two sentences. "...is before other.From(). If they're equal, ..."

josharian (approved):
LGTM. Two unrelated documentation nits, if you're game.

netaddr.go
danderson: Done.

netaddr.go
danderson: Done.

PR #177, opened 2021-05-21
Make IPSetBuilder setters return errors, add Must variants.

### LINKED ISSUE #141: IPSet mysteriously does nothing on invalid inputs
If I AddRange an invalid range, IPSet does nothing, successfully. This is surprising, a more normal idiom in Go would be to have AddRange return an error, and have a MustAddRange that panics if you really want a shorthand.

It does make the builder more annoying to wield, since a bunch of common adders/removers can now fail, but that feels better to me than silently doing something surprising.

danderson:
Fixes #141.

### DIFF
--- a/inlining_test.go
+++ b/inlining_test.go
@@ -42,11 +42,14 @@ func TestInlining(t *testing.T) {
 		return nil
 	})
 	for _, want := range []string{
-		"(*IPSetBuilder).Add",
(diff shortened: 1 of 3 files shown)


danderson:
Converting to draft, since we're discussing in #178 what our style of error handling should be.

josharian:
@danderson I think this is now superseded.

PR #179, opened 2021-05-23
Add selected IP examples (ExampleIP(), Example_IP.Unmap(), etc)

moreati:
These helped me understand the finer details of netaddr.IP. They should get rendered in the docs, like https://pkg.go.dev/inet.af/netaddr#IPSet


example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	ip, err := netaddr.ParseIP("192.0.2.3")
+	if err != nil {
+		panic(err)
+	}
+
+	// netaddr.IP supports comparison using ==
+	fmt.Println(ip == netaddr.IPv4(192, 0, 2, 3))
+
+	// netaddr.IP can be used as a map key
+	hosts := make(map[netaddr.IP]string)
```
josharian, line 23: I'd do this as a single line:

```go
hosts := map[netaddr.IP]string{ip: "example.net}
```

example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	fmt.Println(hosts)
+	// Output:
+	// true
+	// map[192.0.2.3:example.net]
+}
+
+func ExampleIP_Is4() {
+	ipv4 := netaddr.MustParseIP("192.0.2.3")
+	ipv6 := netaddr.MustParseIP("2001:db8::68")
+	ipEmpty := netaddr.IP{}
```
josharian, line 34: Suggest

```go
var zeroIP netaddr.IP
```

and throughout.

example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	ipv4MappedToIPv6 := netaddr.MustParseIP("::ffff:192.0.2.3")
```
josharian, line 35: Suggest `ipMapped` or `ip4in6`, throughout.

example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	fmt.Printf("ipv4.Is4() -> %v\n", ipv4.Is4())
+	fmt.Printf("ipv6.Is4() -> %v\n", ipv6.Is4())
+	fmt.Printf("ipEmpty.Is4() -> %v\n", ipEmpty.Is4())
+	fmt.Printf("ipv4MappedToIPv6.Is4() -> %v\n", ipv4MappedToIPv6.Is4())
+	// Output:
+	// ipv4.Is4() -> true
+	// ipv6.Is4() -> false
+	// ipEmpty.Is4() -> false
+	// ipv4MappedToIPv6.Is4() -> false
+}
```
josharian, line 46: Blank line between functions, please.

example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	fmt.Printf("ipv6.Is4in6() -> %v\n", ipv6.Is4in6())
+	fmt.Printf("ipEmpty.Is4in6() -> %v\n", ipEmpty.Is4in6())
+	fmt.Printf("ipv4MappedToIPv6.Is4in6() -> %v\n", ipv4MappedToIPv6.Is4in6())
+	// Output:
+	// ipv4.Is4in6() -> false
+	// ipv6.Is4in6() -> false
+	// ipEmpty.Is4in6() -> false
+	// ipv4MappedToIPv6.Is4in6() -> true
+}
+func ExampleIP_Is6() {
```
josharian, line 63: There's a lot of duplication between these. I wonder whether they'd be easier to follow as a part of ExampleIP. Something like:

```go
  examples := []struct{
  name string
  ip IP
  }{
    {name: "ipv4", ip: netaddr.MustParseIP("192.0.2.3")},
    // and so on
  }

  for _, example := range examples {
    fmt.Printf("%s: String=%v Is4=%v Is6=%v Is4in6=%v Is Zero=%v\n", example.name, example.ip.String(), /* ... */)
  }
}
```

Maybe go really wild and use a text.tabwriter to make a pretty table. This would be less useful for copy/paste, but easier to see the big picture of what an IP is and what all of its methods mean without having to hop around between examples.

What do you think?

What do you think, Brad/Dave/Matt?

example_test.go
```diff
@@ -10,6 +10,131 @@ import (
...
+	// ipv6.Unmap() -> 2001:db8::68
+	// ipv4MappedToIPv6.Unmap() -> 192.0.2.3
+}
+
+func ExampleIP_WithZone() {
+	ipv4 := netaddr.MustParseIP("192.0.2.3")
+	ipv6 := netaddr.MustParseIP("2001:db8::68")
+	ipv6Zoned := netaddr.MustParseIP("2001:db8::68%eth0")
+
+	fmt.Printf("ipv4.WithZone(\"newzone\") -> %v\n", ipv4.WithZone("newzone"))
```
josharian, line 127: Maybe do

```go
fmt.Printf(`(%v).WithZone("newzone")` -> %v\n, ipv4, ipv4.WithZone("neozone"))
```

Then the output doesn't require you to refer back to what `ipv4` is.

josharian (commented):
Thanks for adding examples! A few thoughts and comments...

example_test.go
moreati: will do

example_test.go
moreati: Can do. Do you think changing `ipv4`, and `ipv6` to `v4IP` and v6IP` would be a good matching change?

Alternatively eliminating the variables entirely has some appeal.

example_test.go
moreati: I think it's useful to have an example against each method, because people reading the docs are much more likely to find the example for method M in situ - next to the description.

I like the idea of a table in the `IP` example, in addition to the individual method examples. It's a slightly higher maintenance burden, for a much higher end user benefit.

example_test.go
moreati: Agree the output should show before and after. Tempted to make this a table as well. I'll try some things and see what sticks.

example_test.go
moreati: Alternatively `var nilIP IP`, as used in

https://github.com/inetaf/netaddr/blob/d57edf19c51753b7d583a5f5054d7d1d2d3cb1ad/netaddr_test.go#L552-L554

josharian: I'd keep ipv4 and ipv6 as-is. I'm ok with zero or zeroIP, or even ipZero. I'd rather not use nilIP.

example_test.go
moreati: `ip4in6` it is

example_test.go
moreati: I've kept the individual method examples, and added a table to `ExampleIP`

example_test.go
```diff
@@ -6,10 +6,161 @@ package netaddr_test
...
+func ExampleIP_IsZero() {
+	var zeroIP netaddr.IP
+	ipv4AllZeroes := netaddr.MustParseIP("0.0.0.0")
+	ipv6AllZeroes := netaddr.MustParseIP("::")
+
+	fmt.Printf("zeroIP.IsZero() -> %v\n", zeroIP.IsZero())
+	fmt.Printf("(%v).IsZero() -> %v\n", ipv4AllZeroes, ipv4AllZeroes.IsZero())
+	fmt.Printf("(%v).IsZero() -> %v\n", ipv6AllZeroes, ipv6AllZeroes.IsZero())
+	// Output:
+	// zeroIP.IsZero() -> true
```
josharian, line 115: I'd suggest `IP{}.IsZero() -> true` here and elsewhere.

josharian (commented):
This looks to me, with minor tweaks, but I'd like Brad or Dave to take a look too. Thanks again.

example_test.go
```diff
@@ -6,10 +6,161 @@ package netaddr_test
...
+	// netaddr.IP supports comparison using ==
+	fmt.Println(ip == netaddr.IPv4(192, 0, 2, 3))
+
+	// netaddr.IP can be used as a map key
+	hosts := map[netaddr.IP]string{ip: "example.net"}
+	fmt.Println(hosts)
+
+	fmt.Println()
+
+	var zeroIP netaddr.IP
```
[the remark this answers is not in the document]
moreati: Oh, I assumed the `Example<thing>` had to match a name in the package under test. Will do.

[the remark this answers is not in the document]
josharian: Yep! Little known tidbit. :) https://golang.org/pkg/testing/#hdr-Examples:

> Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. The suffix must start with a lower-case letter.

josharian:
LGTM. Will merge pending a +1 from @bradfitz, @danderson, or @mdlayher.

moreati:
Thanks, I'll squash my commits then

example_test.go
```diff
@@ -6,10 +6,161 @@ package netaddr_test
...
+	"os"
+	"text/tabwriter"
 
 	"inet.af/netaddr"
 )
 
+func ExampleIP() {
+	ip, err := netaddr.ParseIP("192.0.2.3")
+	if err != nil {
+		panic(err)
```
mdlayher, line 18: Nit: we have MustParseIP, should we just use that?

example_test.go
```diff
@@ -6,10 +6,161 @@ package netaddr_test
...
+
+	// netaddr.IP can be used as a map key
+	hosts := map[netaddr.IP]string{ip: "example.net"}
+	fmt.Println(hosts)
+	// Output:
+	// true
+	// map[192.0.2.3:example.net]
+}
+
+func ExampleIP_properties() {
```
mdlayher, line 32: This is a really nice example, well done.

example_test.go
mdlayher: I like the table example.

mdlayher (approved):
LGTM!

example_test.go
moreati: I purposely chose `ParseIP`, because this is an example of how to use the library in a non-testcase context (e.g. an app). Readers may not reliase that `MustParse` can panic.

josharian:
Thanks again. It's great to have useful examples. I also really like how the table example worked out.

PR #185, opened 2021-05-25
Fix IsLinkLocalUnicast() for Ipv6 addresses >= 0xfe81

### LINKED ISSUE #184: netaddr.IP.IsLinkLocalUnicast() disagrees with net.IP
For the IPv6 address feb0::0
```
netaddr.IP.IsLinkLocalUnicast() == false
net.IP.IsLinkLocalUnicast() == true
```
I believe netaddr.IP is in error, and net.IP may be correct. It's possible both are in error, but I don't think so.

[RFC4291 section 2.4](https://datatracker.ietf.org/doc/html/rfc4291#section-2.4)

> The type of an IPv6 address is identified by the high-order bits of the address, as follows:
> | Address type | Binary prefix  | IPv6 notation | Section |
> |--|--|--|--|
> | ... |
> | Link-Local unicast  | 1111111010     |      FE80::/10    |   2.5.6 |

## netaddr 
`netaddr.IP.IsLinkLocalUnicast()` only checks the first 16 bits equal 0xfeb0, it doesn't mask.
https://github.com/inetaf/netaddr/blob/c0eff8545de6113ee7899372f356c02134f2e02a/netaddr.go#L548

## net
`net.IP.IsLinkLocalUnicast()` (since https://github.com/golang/go/commit/91ba0abef12df6824bcf912ce4bc62f5c9e1fe91 in go 1.5beta2 ) checks the first byte equals `0xfe`, and the second byte masked with `0b1100_0000` == `0b1000_0000`

```go
    return len(ip) == IPv6len && ip[0] == 0xfe && ip[1]&0xc0 == 0x80
```
-- https://github.com/golang/go/blob/f22ec51debeddc0903096e66bfaf641568bede3b/src/net/ip.go#L179

## Analysis

Going purely from RFC4291 section 2.4 I would expect the correct test to be
```go
    return ip.v6u16(0) & 0xffc0 == 0xfe80
```
where
- `0xffc0 == 0b1111_1111_1100_0000` is the netmask
- `0xfe80 == 0b1111_1110_1000_0000` is the prefix

moreati:
Signed-off-by: Alex Willmer <alex@moreati.org.uk>

fixes #184

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,1 +1,2 @@
 Tailscale Inc.
+Alex Willmer <alex@moreati.org.uk>
--- a/netaddr.go
+++ b/netaddr.go
@@ -542,10 +542,10 @@ func (ip IP) IsLinkLocalUnicast() bool {
 	if ip.Is4() {
 		return ip.v4(0) == 169 && ip.v4(1) == 254
 	}
-	// IP Version 6 Addressing Architecture (2.5.6 Link-Local IPv6 Unicast Addresses)
-	// https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.6
+	// IP Version 6 Addressing Architecture (2.4 Address Type Identification)
+	// https://datatracker.ietf.org/doc/html/rfc4291#section-2.4
 	if ip.Is6() {
-		return ip.v6u16(0) == 0xfe80
+		return ip.v6u16(0)&0xffc0 == 0xfe80
 	}
 	return false // zero value
 }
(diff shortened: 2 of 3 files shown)


mdlayher (approved):
https://play.golang.org/p/5NBWu2VOP84

LGTM. I'm so used to `fe80::/64` prefixes that my mind never bothered to do the subnet math for the /10.

@bradfitz @josharian @danderson how are we handling AUTHORS updates? I'm not part of Tailscale and have been contributing for a while. Should we automate it?

josharian:
> @bradfitz @josharian @danderson how are we handling AUTHORS updates? I'm not part of Tailscale and have been contributing for a while. Should we automate it?

I don't think it's worth automating; the volume is small and we're hoping to upstream soon anyway.

That said, @moreati, would you sending the AUTHORS change separately from the bug fix? That way if for any reason we end up reverting the bug fix we don't end up accidentally reverting the AUTHORS change as well? Thanks.

bradfitz:
Merged as e4c51c86b9f85f324005b1c44a4eacf4fd1a5151

PR #186, opened 2021-05-31
Implement changes from design meeting

danderson:
This implements the result of our design discussion last week: IP and IPPort support IPv6 zones, all other types silently strip or ignore the zone.

Also makes IPSetBuilder accumulate and return errors resulting from being given invalid values, rather than silently ignore them.


danderson:
Ran fuzzing for 15min (226M execs, yay 32 hyperthreads), found no new coverage or crashers.

ipset.go
```diff
@@ -237,12 +277,18 @@ func discardf(format string, args ...interface{}) {}
 var debugf = discardf
 
 // IPSet returns an immutable IPSet representing the current state of
-// s. The builder remains usable after calling IPSet.
-func (s *IPSetBuilder) IPSet() *IPSet {
+// s, along with any errors accumulated during the construction of s.
```
mdlayher, line 280: I'm glad this is documented but I also feel like APIs that can return a partially constructed object along with errors are kind of weird to deal with. Are there legitimate reasons why one might want this capability?

ipset.go
```diff
@@ -378,5 +424,16 @@ func (s *IPSet) RemoveFreePrefix(bitLen uint8) (p IPPrefix, newSet *IPSet, ok bo
 	var b IPSetBuilder
 	b.AddSet(s)
 	b.RemovePrefix(prefix)
-	return prefix, b.IPSet(), true
+	newSet, _ = b.IPSet()
```
mdlayher, line 427: I'd feel better about a panic on error here as that means some fundamental invariants have been violated.

ipset_test.go
```diff
@@ -13,6 +13,14 @@ import (
 	"testing"
 )
 
+func buildIPSet(b *IPSetBuilder) *IPSet {
```
mdlayher, line 16: "mustIPSet" or similar to denote the panic?

netaddr_test.go
```diff
@@ -1157,6 +1156,26 @@ func TestIPPrefixMarshalUnmarshal(t *testing.T) {
 	}
 }
 
+func TestIPPrefixMarshalUnmarshalZone(t *testing.T) {
+	orig := `"fe80::1cc0:3e8c:119f:c2e1%ens18/128"`
```
mdlayher, line 1160: Nit: const on both.

mdlayher (approved):
LGTM overall.

netaddr.go
```diff
@@ -1163,7 +1163,12 @@ type IPPrefix struct {
 
 // IPPrefixFrom returns an IPPrefix with IP ip and port port.
```
bradfitz, line 1206: probably worth also mentioning here that zones are stripped.

// If ip has a zone, it is removed. See package docs.

ipset.go
```diff
@@ -158,15 +166,27 @@ func (s *IPSetBuilder) Clone() *IPSetBuilder {
 }
 
 // Add adds ip to s.
-func (s *IPSetBuilder) Add(ip IP) { s.AddRange(IPRangeFrom(ip, ip)) }
+func (s *IPSetBuilder) Add(ip IP) {
+	if ip.IsZero() {
+		return
```
bradfitz, line 171: why doesn't this accumulate an error?

ipset.go
bradfitz: It was so people could ignore this error in the common case, as none of the errors really matter much usually.

But @danderson, I'd document it more explicitly. It does say "along with" but the norm in Go APIs is so strongly "one or the other" that things that return two non-zero values really need to be yelled.

It probably doesn't need to be as verbose as, say, https://golang.org/pkg/io/#Reader but a bit closer.

josharian:
I missed the review window for this, but I'll send my comments etc as PRs.

PR #187, opened 2021-06-02
netaddr: add IPv4Zero

mdlayher:

I need this for work, and it has symmetry with net.IPv4zero and netaddr.IPv6Unspecified.

### DIFF
--- a/inlining_test.go
+++ b/inlining_test.go
@@ -89,6 +89,7 @@ func TestInlining(t *testing.T) {
 		"IPRange.To",
 		"IPRange.From",
 		"IPv4",
+		"IPv4Zero",
 		"IPFrom4",
 		"IPv6LinkLocalAllNodes",
 		"IPv6Unspecified",
--- a/netaddr.go
+++ b/netaddr.go
@@ -93,6 +93,9 @@ func IPv4(a, b, c, d uint8) IP {
 	}
 }
 
+// IPv4Zero returns the IPv4 zero address 0.0.0.0.
+func IPv4Zero() IP { return IP{z: z4} }
+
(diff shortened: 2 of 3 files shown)


bradfitz:
Not a huge fan, as `netaddr.IPv4(0, 0, 0, 0)` is more explicit and shorter than `netaddr.IPv6Unspecified()` too. (https://play.golang.org/p/1jCC_et5dRU)

bradfitz:
Also, `IPv4Zero` sounds like `IP.IsZero`. Shouldn't it be `IPv4Unspecified` to match?

mdlayher:
No strong feelings here. I went with the name the stdlib uses but could also just do an inline zero address as you mention.

bradfitz:
Oh, hah, I didn't even know `net` had that. But it also has both:

```
    IPv6zero                   = IP{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
    IPv6unspecified            = IP{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
```

bradfitz:
Sounds like maybe we don't need this now, with IP.IsUnspecified?

mdlayher:
AFK for a bit so I forget exactly, but I think I don't need it. We do still have IPv6Unspecified. Should we remove that one for the same reason?

bradfitz:
IPv6Unspecified is a bit harder to type by hand, though.

mdlayher:
Nope, don't need this anymore. I do use IPv6Unspecified a lot in tests. Will close.

PR #188, opened 2021-06-02
improve IPSetBuilder error messages

josharian:
Please review commit-by-commit, but note that some later commits refine the earlier ones. (I can squash if desired, but I thought the evolution might be useful and easier to review than a mass rewrite.)

The overarching goal is to make the error messages from IPSetBuilder more useful and actionable.

I'm not particular about any of the details. I tried to make the error messages single-line, because that is my preference (for logging, etc.), but a more verbose multi-line form might be more useful.

PR #189, opened 2021-06-02
document IPSetBuilder API more, make IPSetBuilder.IPSet clear errors

josharian:


mdlayher (approved):
LGTM.

ipset.go
```diff
@@ -287,6 +299,7 @@ func (s *IPSetBuilder) IPSet() (*IPSet, error) {
 	if len(s.errs) == 0 {
 		return ret, nil
 	} else {
+		s.errs = s.errs[:0:0]
```
bradfitz, line 302: This seems wrong. You're clearing it before returning it?

josharian: Oops. Fixed, thanks. The test that would have caught this bug is in #188.

josharian:
I'm going AFK in a few hours. Please feel free to edit and merge at will.

PR #190, opened 2021-06-03
make Contains methods return false when called with an IP method with a zone

josharian:


ipset_test.go
```diff
@@ -528,6 +528,8 @@ func TestIPSetContains(t *testing.T) {
 		{"fc00::1", true},
 		{"fd00::1", true},
 		{"ff00::1", false},
+
+		{"fd00::%a", false},
```
bradfitz, line 532: add `fd00::1%a` too, to contrast with line 529's true

bradfitz (approved):

PR #192, opened 2021-06-22
netaddr: add IP.IsGlobalUnicast

mdlayher:

Another helper I need for CoreRAD. I'd also like to add functions for IPv6 ULA detection, so maybe we can mirror Go 1.17's https://tip.golang.org/pkg/net/#IP.IsPrivate in a follow-up PR.

Notably the new function matches the stdlib logic but since we don't have IPv4zero (pending #187) or IPv4 broadcast, we just declare those inline. It's probably not a big deal though.

I know this has grown organically, but as a matter of tidying, I arrange some table test fields alphabetically and adjusted the comparisons to match. If this is undesirable or noisy, I'm happy to revert and just keep appending fields.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -117,6 +117,46 @@ func ExampleIP_IsZero() {
 	// (::).IsZero() -> false
 }
 
+func ExampleIP_IsGlobalUnicast() {
+	var (
+		zeroIP netaddr.IP
+
+		ipv4AllZeroes = netaddr.MustParseIP("0.0.0.0")
+		ipv4          = netaddr.MustParseIP("192.0.2.3")
+
+		ipv6AllZeroes  = netaddr.MustParseIP("::")
+		ipv6LinkLocal  = netaddr.MustParseIP("fe80::1")
+		ipv6           = netaddr.MustParseIP("2001:db8::68")
+		ipv6Unassigned = netaddr.MustParseIP("4000::1")
+		ip4in6         = netaddr.MustParseIP("::ffff:192.0.2.3")
+	)
+
+	fmt.Printf("IP{}.IsGlobalUnicast() -> %v\n", zeroIP.IsGlobalUnicast())
+
+	ips := []netaddr.IP{
+		ipv4AllZeroes,
+		ipv4,
+		ipv6AllZeroes,
+		ipv6LinkLocal,
+		ipv6,
+		ipv6Unassigned,
+		ip4in6,
+	}
+
+	for _, ip := range ips {
+		fmt.Printf("(%v).IsGlobalUnicast() -> %v\n", ip, ip.IsGlobalUnicast())
+	}
+	// Output:
+	// IP{}.IsGlobalUnicast() -> false
+	// (0.0.0.0).IsGlobalUnicast() -> false
+	// (192.0.2.3).IsGlobalUnicast() -> true
+	// (::).IsGlobalUnicast() -> false
+	// (fe80::1).IsGlobalUnicast() -> false
+	// (2001:db8::68).IsGlobalUnicast() -> true
+	// (4000::1).IsGlobalUnicast() -> true
+	// (::ffff:c000:203).IsGlobalUnicast() -> true
+}
+
 func ExampleIP_String() {
 	ipv4 := netaddr.MustParseIP("192.0.2.3")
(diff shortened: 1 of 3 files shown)


netaddr.go
```diff
@@ -632,6 +632,23 @@ func (ip IP) IsLinkLocalMulticast() bool {
 	return false // zero value
 }
 
+// IsGlobalUnicast reports whether ip is a global unicast address.
+// If ip is the zero value, it will return false.
```
bradfitz, line 636: I think this second sentence can be dropped. That's pretty implicit, no? I don't think we say that for all the other methods where it's also true.

netaddr.go
```diff
@@ -632,6 +632,23 @@ func (ip IP) IsLinkLocalMulticast() bool {
 	return false // zero value
 }
 
+// IsGlobalUnicast reports whether ip is a global unicast address.
```
bradfitz, line 635: for IPv6, is this only 2000::/3 or does it also include ULAs? It should be clear from docs.

netaddr.go
```diff
@@ -632,6 +632,23 @@ func (ip IP) IsLinkLocalMulticast() bool {
...
+// If ip is the zero value, it will return false.
+func (ip IP) IsGlobalUnicast() bool {
+	if ip.z == z0 {
+		// Invalid or zero-value.
+		return false
+	}
+
+	// Match the stdlib's IsGlobalUnicast logic. Notably private IPv4 addresses
+	// and ULA IPv6 addresses are still considered "global unicast".
+	return ip != IPv4(255, 255, 255, 255) &&
```
bradfitz, line 645: maybe `if ip.Is4() {   if v4checks || v4checks { return false } }` first?

netaddr.go
mdlayher: We do say it for the others at the moment, see docs starting at https://pkg.go.dev/inet.af/netaddr#IP.IsInterfaceLocalMulticast.

netaddr.go
mdlayher: Should I use the stdlib comment as a starting point?

> The identification of global unicast addresses uses address type identification as defined in RFC 1122, RFC 4632 and RFC 4291 with the exception of IPv4 directed broadcast addresses. It returns true even if ip is in IPv4 private address space or local IPv6 unicast address space.

netaddr.go
mdlayher: Can do.

mdlayher: How about now?

mdlayher:
@bradfitz ping when you get a sec, thanks!

netaddr.go
bradfitz: That looks like classic Mikioh style documentation, referencing a sea of RFCs numbers & sections. That's fine to include, but I'd also like something more readable with higher-level descriptions/summaries/examples. I shouldn't need to go read a dozen RFCs to understand such a simple-looking function.

mdlayher: @bradfitz updated, feedback welcome. It almost seems like this might be better as a bulleted list, but what do you think?

josharian:
I don't have many opinions here, but if this goes in, please add it to the examples, mirroring the other IP properties. Thanks.

mdlayher:
@josharian sure, would you suggest a new top-level example function? I think adding it to the existing _properties example could result in information overload with the table output.

josharian:
A new top level example for it, in the form of a table, with an illustrative set of IPs (maybe with comments?) would probably help a lot.

I'm 50/50 on also adding it to the existing example. Tables can support a lot of data, and people might wonder about the omission. Make a judgment call.

example_test.go
```diff
@@ -117,6 +117,46 @@ func ExampleIP_IsZero() {
 	// (::).IsZero() -> false
 }
 
+func ExampleIP_IsGlobalUnicast() {
```
mdlayher, line 120: @josharian added example with some relevant addresses, PTAL.

bradfitz (approved):
If this both matches std and nobody else objects, I'm fine with this now with its new docs.

mdlayher:
Thanks! I know there are a couple issues open for checking if we agree with stdlib but no existing test code for that AFAIK. I know there has been some work in PRs for that.

mdlayher:
Seeing no objections, will merge.

PR #196, opened 2021-07-03
netaddr: add IP.IsUnspecified

### LINKED ISSUE #195: add IP.Unspecified?
I was just porting some code from net.IP to netaddr.IP and discovered we don't have netaddr.IP.IsUnspecified like std.

Add it?

/cc @danderson @josharian @mdlayher

mdlayher:

I need this too, so I whipped it up. PTAL.

Fixes #195 .

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -402,7 +402,8 @@ func (ip IP) v6u16(i uint8) uint16 {
 // IsZero reports whether ip is the zero value of the IP type.
 // The zero value is not a valid IP address of any type.
 //
-// Note that "0.0.0.0" and "::" are not the zero value.
+// Note that "0.0.0.0" and "::" are not the zero value. Use IsUnspecified to
+// check for these values instead.
 func (ip IP) IsZero() bool {
 	// Faster than comparing ip == IP{}, but effectively equivalent,
 	// as there's no way to make an IP with a nil z from this package.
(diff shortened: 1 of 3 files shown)


netaddr.go
```diff
@@ -659,6 +660,22 @@ func (ip IP) IsGlobalUnicast() bool {
 		!ip.IsLinkLocalUnicast()
 }
 
+// IsUnspecified reports whether ip is an unspecified address, either the IPv4
+// address "0.0.0.0" or the IPv6 address "::".
+//
+// Note that the ip zero value is not an unspecified address. Use IsZero to
```
bradfitz, line 666: IP zero value

bradfitz (approved):

netaddr.go
```diff
@@ -659,6 +660,22 @@ func (ip IP) IsGlobalUnicast() bool {
...
+// check for the zero value instead.
+func (ip IP) IsUnspecified() bool {
+	if ip.Is4() {
```
bradfitz, line 669: did you write it like this for performance or for clarity?

Because I'd think the one-liner would be the same speed & clarity:

    return ip == IPv4(0,0,0,0) || ip == IPv6Unspecified()

is it not?

In any case, I don't care much how it's written exactly. Just curious.

mdlayher: Clarity and consistency with other methods. No preference here.

mdlayher: Changing to simple return with ||.

PR #202, opened 2021-07-16
netaddr: add functionality to IPRange for feature parity

### LINKED ISSUE #201: Add methods to IPRange for parity with other types
The `IPRange` lack some functions and methods that consistently exist on`IP`, `IPPort`, and `IPPrefix`. See the following table.

| `func` | `IP` | `IPPort` | `IPPrefix` | `IPRange` |
| --- | --- | --- | --- | --- |
| `func ParseT(string) (T, error)` | ✅ | ✅ | ✅ | ✅ |
| `func MustParseT(string) T` | ✅ | ✅ | ✅ | ❌ |
| `func (T) IsValid() bool` | ✅ | ✅ | ✅ | ✅ |
| `func (T) IsZero() bool` | ✅ | ✅ | ✅ | ❌ |
| `func (T) String() string` | ✅ | ✅ | ✅ | ✅ |
| `func (T) AppendTo([]byte) []byte` | ✅ | ✅ | ✅ | ❌ |
| `func (T) MarshalText() ([]byte, error)` | ✅ | ✅ | ✅ | ❌ |
| `func (*T) UnmarshalText([]byte) error` | ✅ | ✅ | ✅ | ❌ |

Given that a parse method exists already for `IPRange`, I don't see any reason why we can't also provide `MustParse`, `AppendTo`, `MarshalText`, and `UnmarshalText`.

The use case of this change is for [tailscale.com/util/deephash](https://pkg.go.dev/tailscale.com/util/deephash), where the `netaddr.IPRange` type is the only type ever visited by `deephash` where the package must touch unexported fields because the type has no `AppendTo` method (which `deephash` [specially understands](https://github.com/tailscale/tailscale/blob/171ec9f8f4e4b4751b5f07dbed6f01514597c1a0/util/deephash/deephash.go#L115)). Adding `AppendTo` would be a one step closer to generating `deephash` functionality for all relevant tailscale types.

\cc @bradfitz @josharian

dsnet:
Add MustParseIPRange, IPRange.{IsZero,AppendTo,MarshalText,UnmarshalText}
to bring it to feature parity with the other common types.

Similar to the MarshalText methods of the other types,
it never returns an error for invalid values.
This behavior should probably change, but should be consistently
done for all other types (see #169).

Fixes #201


bradfitz (approved):

danderson:
Missing signed-off-by

bradfitz (approved):

PR #204, opened 2021-07-16
netaddr: fix fuzz build

dsnet:
The appendMarshaller type is declared in both fuzz.go and netaddr_test.go,
which causes the fuzz tests to fail to build. Use the one in netaddr_test.go.
While we're at it, rename it to appendMarshaler.


bradfitz (approved):

mdlayher (approved):
LGTM, needs DCO too.

PR #206, opened 2021-07-19
netaddr: remove unneeded named/bare returns

mdlayher:

Bare returns are too magic, and IMO the named returns I've removed here don't have much value. The only exported one which I think might be justified is https://pkg.go.dev/inet.af/netaddr#IPSet.RemoveFreePrefix, which I've left alone.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -175,7 +175,7 @@ func (err parseIPError) Error() string {
 }
 
 // parseIPv4 parses s as an IPv4 address (in form "192.168.0.1").
-func parseIPv4(s string) (ip IP, err error) {
+func parseIPv4(s string) (IP, error) {
 	var fields [3]uint8
 	var val, pos int
 	for i := 0; i < len(s); i++ {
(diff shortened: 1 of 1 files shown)


netaddr.go
```diff
@@ -359,7 +359,7 @@ func parseIPv6(in string) (IP, error) {
 // the standard library's ParseIP: https://play.golang.org/p/qdjylUkKWxl.
 // To convert a standard library IP without the implicit unmapping, use
 // FromStdIPRaw.
-func FromStdIP(std net.IP) (ip IP, ok bool) {
```
bradfitz, line 362: please restore the names here. The bool is mysterious otherwise, and the docs then refer to a mysterious "ok" from line 354.

netaddr.go
```diff
@@ -371,7 +371,7 @@ func FromStdIP(std net.IP) (ip IP, ok bool) {
 // If std is invalid, ok is false.
 // Unlike FromStdIP, FromStdIPRaw does not do an implicit Unmap if
 // len(std) == 16 and contains an IPv6-mapped IPv4 address.
-func FromStdIPRaw(std net.IP) (ip IP, ok bool) {
+func FromStdIPRaw(std net.IP) (IP, bool) {
```
bradfitz, line 374: same

bradfitz (changes requested):

netaddr.go
mdlayher: I'm happy to update the docs accordingly, but isn't the comma-ok idiom common enough in Go where the meaning can be reasonably inferred without the names? Same for below.

mdlayher: Really I guess these statements would apply to all the changes I've made here.

bradfitz: I think 1 or 2 lines of this PR were fine.

mdlayher:
Ended up never following up on this, closing.

PR #207, opened 2021-07-19
netaddr: add test for IP.UnmarshalBinary with bad byte length

mdlayher:

tests++

### DIFF
--- a/netaddr_test.go
+++ b/netaddr_test.go
@@ -363,9 +363,17 @@ func TestIPMarshalUnmarshalBinary(t *testing.T) {
 	}
 
 	// Cannot unmarshal into a non-zero IP
-	ip := MustParseIP("1.2.3.4")
-	if err := ip.UnmarshalBinary([]byte{1, 1, 1, 1}); err == nil {
+	ip1 := MustParseIP("1.2.3.4")
+	if err := ip1.UnmarshalBinary([]byte{1, 1, 1, 1}); err == nil {
 		t.Fatal("unmarshaled into non-empty IP")
+	}
+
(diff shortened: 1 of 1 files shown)


netaddr_test.go
```diff
@@ -363,10 +363,16 @@ func TestIPMarshalUnmarshalBinary(t *testing.T) {
...
 
 	// Cannot unmarshal into a non-zero IP
-	ip := MustParseIP("1.2.3.4")
-	if err := ip.UnmarshalBinary([]byte{1, 1, 1, 1}); err == nil {
+	ip1 := MustParseIP("1.2.3.4")
+	if err := ip1.UnmarshalBinary([]byte{1, 1, 1, 1}); err == nil {
 		t.Fatal("unmarshaled into non-empty IP")
 	}
+
+	// Cannot unmarshal from unexpected IP length.
```
dsnet, line 371: Should there be  a test for unmarshaling from too long of a buffer?

dsnet (approved):

netaddr_test.go
mdlayher: Sure, done!

mdlayher: Oh sorry, did you mean >16 bytes? If so that's covered by the IPv6+zone cases above. But will see if you were thinking the same thing I was before reverting the loop.

dsnet: 3B and 5B were the cases I was thinking of.

mdlayher: Nice, any objections to the updated code?

dsnet: None from me.

PR #219, opened 2022-02-08
Add dotted decimal output for mapped addr

cmol:
For IPv4 mapped IPv6 addresses it is commont to see the IPv4 part of the
address as dotted decmal after a prefix of `::ffff:` making
`::ffff:c000:0202` be printed as `::ffff:192.0.2.2`.

This commit implements a method that prints this format for mapped
addresses, and uses `String()` for all non mapped addresses.

### DIFF
--- a/netaddr.go
+++ b/netaddr.go
@@ -953,6 +953,28 @@ func (ip IP) appendTo6(ret []byte) []byte {
 	return ret
 }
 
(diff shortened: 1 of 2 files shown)


bradfitz:
Now that this is in Go 1.18, we're no longer adding features to this package.

Also, this is already the behavior in Go 1.18: https://go.dev/play/p/mexPHj_mn8a?v=gotip

